PR #56, opened 2022-12-08
Warg client skeleton

Kylebrown9:
Implement skeleton of warg client

### DIFF
--- a/crates/warg-client/src/lib.rs
+++ b/crates/warg-client/src/lib.rs
@@ -1,58 +1,114 @@
-use std::collections::HashMap;
+use semver::Version;
+use url::Url;
 
-struct Client<Cache> {
-    cache: Cache
+/// A client for interacting with a registry
+#[allow(dead_code)]
+pub struct Client<RegistryCache, ValidationCache> {
+    registry: DomainName,
+    known_roots: Vec<Root>,
+    registry_cache: RegistryCache,
+    validation_cache: ValidationCache,
 }
 
-struct Domain(String);
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DomainName(String);
 
-impl Client {
-    pub fn new(primary_registry: Domain) -> Self {
-        todo!()
+#[allow(dead_code)]
+pub struct Root {
+    bytes: Vec<u8>
+}
+
+pub trait Cache {
+    /// Insert a value into the clients cache.
+    fn insert(&mut self, key: Vec<u8>, value: Vec<u8>);
+
+    /// Lookup a previously inserted cache value.
+    fn lookup(&self, key: &[u8]) -> Option<Vec<u8>>;
+}
+
+impl<RegistryCache, ValidationCache> Client<RegistryCache, ValidationCache>
+where
+    RegistryCache: Cache,
+    ValidationCache: Cache
+{
+    /// Create a new client with no previous state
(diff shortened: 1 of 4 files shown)


crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    known_roots: Vec<Root>,
+    registry_cache: RegistryCache,
+    validation_cache: ValidationCache,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DomainName(String);
+
+#[allow(dead_code)]
+pub struct Root {
```
lann, line 17: Maybe a more descriptive name (if it needs to be `pub`)?

crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    bytes: Vec<u8>
+}
+
+pub trait Cache {
+    /// Insert a value into the clients cache.
+    fn insert(&mut self, key: Vec<u8>, value: Vec<u8>);
+
+    /// Lookup a previously inserted cache value.
+    fn lookup(&self, key: &[u8]) -> Option<Vec<u8>>;
```
lann, line 26: Probably better to return `Option<&[u8]>` and let the consumer clone if necessary.

crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
+use semver::Version;
+use url::Url;
+
+/// A client for interacting with a registry
+#[allow(dead_code)]
+pub struct Client<RegistryCache, ValidationCache> {
+    registry: DomainName,
```
lann, line 7: Any particular reason to configure the registry by just domain name vs URL?

crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        self.registry.clone()
+    }
+
+    /// Serialize the client state
+    pub fn to_state(&self) -> Vec<u8> {
+        unimplemented!()
+    }
+
+    /// Construct a registry client with prior state
+    pub fn from_state(registry: DomainName, state: Vec<u8>, registry_cache: RegistryCache, validation_cache: ValidationCache) -> Self {
```
lann, line 55: Seems like you might already be getting into enough complexity here that a `ClientBuilder` might make sense.

crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub name: String,
+    /// What kind of package this is
+    pub package_type: PackageType,
+    /// The available releases of the package.
+    /// Does not include yanked releases.
+    pub releases: Vec<ReleaseInfo>
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PackageType {
```
lann, line 91:
```suggestion
#[non_exhaustive]
pub enum PackageType {
```

crates/warg-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// The release content digest
+    pub digest: Vec<u8>,
+    /// The known locations where the content may be found
+    pub known_locations: Vec<Location>
+}
+
+/// A way of retrieving package contents
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Location {
```
lann, line 112: I thought about this a bit when looking at the protocol and I don't _think_ we want this to be an enum, at least at this layer.

PR #59, opened 2023-01-05
Add warg-server crate

lann:
This commit adds a Axum-based HTTP server which currently just has simple content upload and download handlers for demo purposes.


Kylebrown9 (approved):
Looks good

PR #61, opened 2023-01-06
Simplify package log validation.

peterhuene:
This PR attempts to simplify package log validation to make it easier to use.

Currently validation is driven by multiple enum state machines and it makes it a bit cumbersome to use as an API.

The validation machinery has been replaced by a single `Validator` type which is fed package records via a `validate` method; the validation state is kept internally to the validator, but it can be queried relating to package releases and public keys. The release information has been beefed up to make it easier to determine who released/yanked a particular version as well.

Additionally, the validator is serializable so it's now possible to save the current validation state and rehydrate the validator where it left off processing records. The intention behind this change is to (optionally) skip over previously validated entries when querying a package log for release information.

For example, `cargo-component` will shortly support local (file system) component registries that users can publish local components or vendor packages from remote registries. It will usually want to save the validation state of the package logs for faster querying and appending to the log, but will do a full validation if the state fails to deserialize or explicitly requested by the user (e.g. a `--validate` CLI option to throw out the previous validator state and scan the entire log).

The package log tests have also been updated to make it easier to "bless" known-good output baselines, similar to what we do in other BCA repo tests.

Removed some unnecessary cloning in places.

Some minor formatting and source layout changes are included, but that's a preference thing and can be reverted if desired.

Lastly, if we like and merge this `Validator` approach, I will follow-up with an additional PR for operator logs.


crates/warg/src/package/model.rs
```diff
@@ -20,13 +18,20 @@ pub struct PackageRecord {
 }
 
 /// Each permission represents the ability to use the specified entry
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Serialize, Deserialize)]
```
lann, line 21: Maybe `#[serde(rename_all = "lowercase")]` to match `Display`?

crates/warg/tests/package.rs
```diff
@@ -1,81 +1,105 @@
...
+};
 use warg_protocol::{
     hash,
-    package::{self, validate::ValidationState},
+    package::{self, validate::Validator},
     protobuf, signing, Envelope,
 };
 
 #[test]
-fn test_package_logs() {
```
Kylebrown9, line 15: I get using `?` early returns in main code, but in a test I usually like that `unwrap` or `expect` will bail where it failed and give me a convenient trace.

crates/warg/tests/package.rs
```diff
@@ -1,81 +1,105 @@
...
-            record.prev = last.clone();
 
-            let envelope = Envelope::signed_contents(&key, record).unwrap();
+    Ok(())
+}
 
-            *last = Some(hash::HashAlgorithm::Sha256.digest(envelope.content_bytes()));
+fn validate_input(input: Vec<EnvelopeData>) -> Result<Validator> {
+    let mut validator = Validator::new();
+    for record in input.into_iter().scan(None, |last, e_data| {
```
Kylebrown9, line 33: I think with a `scan` this big, it'll look better to chain it with a `fold` than make it the right-hand side of a loop.

crates/warg/tests/package.rs
```diff
@@ -1,81 +1,105 @@
...
+        })?;
+
+    let output = match validate_input(input) {
+        Ok(validator) => Output::Valid(validator),
+        Err(e) => Output::Error(e.to_string()),
     };
 
-    assert_eq!(test.output, result);
-}
+    if std::env::var_os("BLESS").is_some() {
```
Kylebrown9, line 76: I'm planning to add two more features to the test infrastructure and I'm curious how you would combine them with the changes you've made here like `BLESS`.
1. Lettings test-writers use aliases (e.g. `{alice-pub}`) in their JSON (for input and output) that are replaced with their actual values from some central file before deserialization.
2. Having the test json generate encoded protobuf that lives in the repository so that we know that a proto we encoded in the past still works instead of / in addition to knowing that the proto we encode today works.

crates/warg/src/package.rs
```diff
@@ -1,15 +1,13 @@
...
-use crate::hash;
-
 pub mod model;
 pub mod validate;
 
-use crate::{protobuf, Decode, Encode, Signable};
-
-// Deserialization
+/// The currently supported package protocol version.
+pub const PACKAGE_RECORD_VERSION: u32 = 0;
```
Kylebrown9, line 10: If we're going to add a version constant for package records, operator records should get one too.

crates/warg/src/package/validate.rs
```diff
@@ -71,62 +70,76 @@ pub enum ValidationError {
...
 
-#[allow(clippy::large_enum_variant)]
-#[derive(Default, Debug, Clone, PartialEq, Eq)]
-pub enum ValidationState {
-    #[default]
-    Uninitialized,
-    Initialized(ValidationStateInit),
+/// Represents information about a package yanking.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Yank {
```
Kylebrown9, line 75: I think since the entry for yanking is also just called `Yank` (albeit with enum prefix), this should maybe be called `YankInfo` or `YankContext`.

crates/warg/src/package/validate.rs
```diff
@@ -191,267 +251,262 @@ impl ValidationState {
...
+            {
+                self.validate_init_entry(signer_key_id, *hash_algorithm, key)?;
+                continue;
+            }
 
-        match entry_validation_state {
-            Some(state) => Ok(state),
-            None => Err(ValidationError::InitialRecordDoesNotInit),
+            // Must have seen an init entry by now
+            if self.algorithm.is_none() {
```
Kylebrown9, line 285: Consider adding `initialized(&self) -> bool` to use places where we want to know if initialized to make it more explicit.

crates/warg/src/lib.rs
```diff
@@ -157,3 +152,49 @@ pub trait Decode: Sized {
 pub trait Encode {
     fn encode(&self) -> Vec<u8>;
 }
+
+/// Helper module for serializing and deserializing timestamps.
+///
+/// This is used over serde's built-in implementation to produce cleaner timestamps
+/// in serialized output.
+mod timestamp {
```
Kylebrown9, line 160: Is this actually used somewhere? Afaik, the only timestamps we serialize use the `pbjson_build` implementation of serialize on the pbjson/prost timestamp type(s).

Kylebrown9 (changes requested):
Overall, it looks good but there are some things I want you to look at before this goes in. Beyond the things I've left as review comments there are two things that didn't have a specific spot that I'll mention here.
1. It looks like the change is more significant for package logs than it would be for operator logs, but that if it makes sense to apply to one of them it should make sense to apply to both, and that would be good for consistency.
2. The previous `ReleaseState` enum in the validator "forgot" about the content associated with yanked releases intentionally. Not including it in the output is my way of saying "this doesn't exist anymore" and getting people to actually respect yank.

crates/warg/src/package.rs
peterhuene: Agreed. I was waiting on touching the operator log code to see if we even wanted to take changes to the package log validation before doing so. I'll have a follow-up for that.

crates/warg/src/lib.rs
peterhuene: It's used in the `Validator` serialization to give a nicer-looking timestamp in the output (`serde_json` serializes `SystemTime` as a struct with verbose names by default).

crates/warg/tests/package.rs
peterhuene: For #1, if it's desirable to have the aliasing syntax in the baseline, would it be possible post-process the serialized output (i.e. as a `serde_json::Value`) to do a reverse look-up of public key to alias? If that's too much, I'm fine with not using a "bless this baseline" scheme.

For #2, I would imagine a test that reads in and validates protobuf package log; we could then simply save a textual representation of the log (json-serialized protobuf envelopes?) next to it as a baseline to compare against, with perhaps the serialized contents of the validator that was used?

crates/warg/src/lib.rs
Kylebrown9: I see, and since `Validator` doesn't have a proto, its allowed to use `SystemTime` directly and not the `google.protobuf.Timestamp` type which has a defined JSON mapping generated by `pbjson_build`.

Kylebrown9 (commented):
One very simple change away from merge-ready.

crates/warg/src/package/validate.rs
```diff
@@ -1,18 +1,17 @@
...
 
-use crate::hash;
-use crate::signing;
-
-use super::model;
-use crate::{Envelope, Signable};
-
 #[derive(Error, Debug)]
 pub enum ValidationError {
+    #[error("The validation state is unsupported")]
```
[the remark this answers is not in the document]
peterhuene: Ah, thanks! Yeah, something added from a very early attempt at these changes. I'll remove.

Kylebrown9 (approved):

PR #63, opened 2023-01-20
Implement Initial Warg Server

Kylebrown9:


crates/warg-crypto/src/hash/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::{fmt, str::FromStr};
+
+mod dynamic;
+mod r#static;
+
+pub use digest::{Digest, Output};
+pub use sha2::Sha256;
+
+pub use dynamic::{DynHash, DynHashParseError};
+pub use r#static::Hash;
```
lann, line 12: :neutral_face:

lann: `r#static`. I think raw idents should mostly be reserved for things like codegen.

Kylebrown9: I think it's not so bad since it's not `pub`, but I will change it if we can come up with a naming scheme that doesn't require it.

crates/warg-server/src/services/core.rs
```diff
@@ -0,0 +1,414 @@
...
+        checkpoint: Arc<Envelope<MapCheckpoint>>,
+    },
+    Rejected {
+        reason: String,
+    },
+}
+
+pub struct CoreService {
+    mailbox: mpsc::Sender<Message>,
+    handle: JoinHandle<State>,
```
lann, line 81: If the goal is to persist to a db anyway, maybe strip some of this graceful shutdown complexity?

Kylebrown9:
This PR has a couple issues I realized with doing a typical fetch-proof flow.
1. Getting records for the latest checkpoint
   * There's no endpoint for getting the current `SerdeEnvelope<MapCheckpoint>`
   * Fetch doesn't let you request records "up to" a certain `MapCheckpoint`
   * Fetch doesn't give you operator records
2. Proof validation
   * It may be too confusing that proof bundles lose information about "what" they prove and you have to know what order you asked for the proofs in to interpret them.
3. Initialization
   * We need to provide a way to initialize the registry with operator contents, which will also need to include them in the transparency structures

Item `1` are things that are just awkward to do without adding more auxiliary data structures (but will be easy when we have a DB) and we probably need to bite the bullet and add them for now to get the semantics we need.

Item `2` is partially a question, should proofs include the base hash being proven to improve correlation or is compactness better?

Item `3` is just more work I need to get to.

Kylebrown9:
As of now, I think the items I previously mentioned are resolved and it should be possible to perform an actual demonstration.

lann (approved):

PR #69, opened 2023-02-13
Misc updates

lann:
- Fix clippy lints
- Rename `forrest` -> `warg-transparency`
- Remove warg- prefix from crate directories, matching `wasmtime` and `wit-bindgen` style


crates/warg-transparency/src/map/map.rs
```diff
@@ -113,7 +113,7 @@ where
     fn clone(&self) -> Self {
         Self {
             link: self.link.clone(),
-            unknown_field: self.unknown_field.clone(),
```
Kylebrown9, line 116: I believe this field is actually just the `len` of the map, I just didn't know what it did when I was refactoring something and gave it this name as a placeholder.

Kylebrown9 (approved):
lgtm

PR #70, opened 2023-03-02
Move the `cli` crate to top level.

peterhuene:
This PR moves and refactors the `warg-cli` crate to be a top-level package in the workspace; this allows users to run the CLI with `cargo run`.

Additionally, the crate is now a library with a `warg-cli` bin, allowing third parties to directly use the CLI commands from another crate.

The crate has been heavily refactored to separate the commands into their own types to encapsulate their execution logic.

The crate has been made to deny missing comments so that the CLI help is fully populated.

Note: the `set-registry` command has been renamed to `init`.


peterhuene:
I plan on updating the root README soon to fill it out to match what we use for other BCA repos, too.

Kylebrown9 (approved):
Thanks for doing this!

PR #72, opened 2023-03-08
Refactor API types into a `warg-api` crate.

peterhuene:
This PR the serializable API types from the server crate into a
`warg-api` crate.

The client and server now share these types via the `warg-api` crate.

This change allows crates dependent upon the `warg-client` crate to not have to
build server-related dependencies.

### DIFF
--- a/crates/api/src/content.rs
+++ b/crates/api/src/content.rs
@@ -1,0 +1,22 @@
+//! Types relating to the content API.
+
+use serde::{Deserialize, Serialize};
+use warg_crypto::hash::DynHash;
+
+/// Represents a content source.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ContentSource {
+    /// The digest of the content.
+    pub digest: DynHash,
+    /// The kind of content source.
+    pub kind: ContentSourceKind,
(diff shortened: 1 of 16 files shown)


crates/api/src/content.rs
```diff
@@ -0,0 +1,22 @@
+//! Types relating to the content API.
+
+use serde::{Deserialize, Serialize};
+use warg_crypto::hash::DynHash;
+
+/// Represents a content source.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
Kylebrown9, line 8: Why the switch to camel case?

peterhuene: Given it's a JSON-serialized format, I figured it'd be more natural to default to with these types (most places are a singular word, so it doesn't impact them, but setting it everywhere guards against the future).

Happy to revert if that doesn't make sense to do.

Kylebrown9: I don't have a strong opinion here, and just picked a default option I tend to like.

Looking at the OCI distribution spec, they're using camel case.
Being consistent with them (and presumably many others) seems like a good enough reason to me.

Kylebrown9 (approved):
LGTM

PR #74, opened 2023-03-10
add info command

macovedj:
Adds command analogous to `yarn info`, printing versions and other stats about a package.
Initially I'm just enabling the printing of the version numbers that exist for a given package, but could easily add any other info that anybody identifies as significant.

### DIFF
--- a/crates/api/src/content.rs
+++ b/crates/api/src/content.rs
@@ -15,7 +15,7 @@ pub struct ContentSource {
 
 /// Represents the supported kinds of content sources.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "camelCase")]
+#[serde(tag = "type", rename_all = "camelCase", content = "c")]
 pub enum ContentSourceKind {
     /// The content is located at an anonymous HTTP URL.
     HttpAnonymous(String),
--- a/crates/api/src/content.rs
+++ b/crates/api/src/content.rs
@@ -15,8 +15,11 @@ pub struct ContentSource {
 
 /// Represents the supported kinds of content sources.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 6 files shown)


crates/api/src/content.rs
```diff
@@ -15,7 +15,7 @@ pub struct ContentSource {
 
 /// Represents the supported kinds of content sources.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "camelCase")]
+#[serde(tag = "type", rename_all = "camelCase", content = "c")]
```
peterhuene, line 18: Rather than use this type of tagged enum, I think we should just go back to having `HttpAnonymous` be a struct -like variant with a `url` field.

Sorry about breaking this when I did the api crate refactoring.

Kylebrown9: I missed this change in your refactor. Agree that we should bring back the struct-like variant and url field. Other than that this PR lgtm.

macovedj: Ah no worries about the change.  Pushed a commit to bring it back.

Kylebrown9 (approved):
Looks good, thanks!

src/commands/info.rs
```diff
@@ -0,0 +1,24 @@
...
+use clap::Args;
+
+/// Update all local packages in the registry.
+#[derive(Args)]
+pub struct InfoCommand {
+    /// The common command options.
+    #[clap(flatten)]
+    pub common: CommonOptions,
+
+    /// The name of the package to install.
```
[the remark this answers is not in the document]
macovedj: Good catch with these!  Thanks!

Kylebrown9:
Good catch @fibonacci1729. @macovedj you can merge when ready.

macovedj:
> Good catch @fibonacci1729. @macovedj you can merge when ready.

Sweet.  I don't think I have merge permissions though.

Kylebrown9:
Forgot it was set up that way. Let's talk about merge policy on Monday.

PR #77, opened 2023-03-16
Expand what CI does.

peterhuene:
This PR updates CI to test on Linux, macOS, and Windows.

It also adds a check to ensure the warg CLI can be installed.


Kylebrown9 (approved):
LGTM

PR #88, opened 2023-04-05
Protocol Wasm

macovedj:
There's a qwik app in here, which likely should live in another repo.  `protocol-wasm` is a new crate created using `cargo-component` that is a copy of the logic in `protocol`.  This may also need to live in another repo until `cargo-component` integrates with workspaces better.  The javascript qwik app uses the generated wasm component to validate package logs and spit out info.  After determining how the validation function will interact with inclusion proofs, we can use origin private file system (opfs) to store package info returned by validations, which can then be used to render info about packages in the app.  Lots of cleanup is still needed


Kylebrown9:
The name "Proto comp" sounds too close to protobuf compiler / protoc for a project that also uses that tool. I think `protocol-wasm` or `protocol-guest` would make sense but I'm open to other options.

PR #89, opened 2023-04-05
Some minor code cleanup.

peterhuene:
* Removed `url` field from `PackageDownload`.
* Made the reqwest client a field of API client.


peterhuene:
The API client change is motivated by cargo component, which is going to use the same warg client across multiple spawned tasks for downloading packages (one warg client per registry being communicated with); doing it this way allows for better connection reuse.

Kylebrown9 (approved):
Thanks! Makes sense to me.

PR #91, opened 2023-04-06
Improve fetch API error message when fetching logs for nonexistent package.

peterhuene:
This commit improves the error message displayed when fetching a log for a nonexistent package.

Previously it would print out the log id of the missing package, which isn't very friendly.

This commit allows the fetch API to return an error of the package name, specifically for fetching logs.


fibonacci1729 (approved):
👍

PR #93, opened 2023-04-10
protocol-wasm

macovedj:
This creates a new workspace for component crates, which includes a component crate for the protocol crate.


components/protocol-wasm/world.wit
```diff
@@ -0,0 +1,67 @@
...
+      content-bytes: list<u8>,
+      key-id: string,
+      signature: string
+    }
+    enum hash-algorithm {
+      sha256
+    }
+    record dyn-hash {
+      algo: hash-algorithm,
+      bytes: option<list<u8>>
```
Kylebrown9, line 14: option?

Cargo-component.lock
Kylebrown9, line 1: What's the difference between the root and `components/` lock files?

components/Cargo.toml
Kylebrown9, line 1: Presumably we need the separate workspace due to some detail about how cargo-component works?

components/protocol-wasm/config.yml
Kylebrown9, line 1: what's this for?

components/protocol-wasm/world.wit
```diff
@@ -0,0 +1,67 @@
...
+    }
+    variant record-id {
+      dyn-hash(dyn-hash)
+    }
+    record head {
+      digest: record-id,
+      timestamp: option<string>
+    }
+    enum perm {
```
Kylebrown9, line 23: permission is not abbreviated throughout rest of the repo

Cargo-component.lock
peterhuene: The latter are lock files from `cargo-component`, which, with the merger of my [current PR to add support for warg registries](https://github.com/bytecodealliance/cargo-component/pull/69), will have store information about how warg registry dependencies were resolved along with their content hashes.

These files will likely be empty for the foreseeable future until these components depend on anything from a warg registry.

components/Cargo.toml
peterhuene: Correct, it's a limitation on how targeting works with cargo. There's an unstable feature to build particular packages to specified targets, but without that, the entire workspace has to build for the same targets.

Thus if you use `cargo component build --all`, every crate in the workspace will build for a wasm target.

For now we'll have a workspace for native warg packages and a workspace for the components.

Cargo-component.lock
Kylebrown9: I get the general idea of why we have `Cargo-component.lock` files. I was just wondering why we have one in both the root and in the `components/` folder and what the difference between those two is.

peterhuene: Oh sorry, I misunderstood your question.

Indeed, I'd expect only one next to the workspace for the components.

components/protocol-wasm/world.wit
macovedj: removed

Cargo-component.lock
macovedj: Got rid of the one in the root

components/protocol-wasm/world.wit
macovedj: Stopped abbreviating

components/protocol-wasm/config.yml
macovedj: deleted

components/protocol-wasm/src/lib.rs
```diff
@@ -0,0 +1,213 @@
+use chrono::{DateTime, Utc};
+use base64::{Engine as _, engine::{self, general_purpose}};
+use std::str::FromStr; // 0.4.15
+
+use bindings::protocol;
```
Kylebrown9, line 5: maybe alias `bindings::protocol` to something (just `bindings` maybe?) so it's clearer that it comes from the component bindings and doesn't refer to the protocol crate itself.

components/protocol-wasm/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+        Self {
+            name: name.into(),
+            checkpoint: None,
+            state: package::Validator::default(),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct MyBody(protocol::ProtoEnvelopeBody);
```
Kylebrown9, line 51: What's this newtype for?

components/protocol-wasm/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+                permissions,
+                releases,
+                keys: Some(keys),
+            },
+            heads
+        };
+    }
+}
+
+bindings::export!(Component);
```
Kylebrown9, line 213: Is putting the bindings gen here a convention? I'd expect it to be as early in the file as rustfmt will allow.

components/protocol-wasm/world.wit
Kylebrown9, line 1: you should be able to define the interface on its own at the top of the file and then make the world just export that interface to avoid nesting all the types twice.

crates/crypto/src/hash/dynamic.rs
Kylebrown9, line 1: Did everything get `pub`-ified for bindings reasons? I'd like to avoid making everything `pub` as it makes it easier to exercise the library incorrectly.

crates/server/src/api/proof.rs
```diff
@@ -74,6 +76,7 @@ async fn prove_consistency(
     State(config): State<Config>,
     Json(body): Json<ConsistencyRequest>,
 ) -> Result<Json<ConsistencyResponse>, ProofApiError> {
+    println!("MADE IT TO HANDLER");
```
Kylebrown9, line 79: Remove

components/protocol-wasm/src/lib.rs
macovedj: How about `use bindings::protocol as protocolbindings`?  I don't think I can use `bindings` as wit-bindgen uses that as a keyword in an export at the bottom, and I also run into issues using `-` and `_` in the alias name

crates/crypto/src/hash/dynamic.rs
macovedj: Yeah exactly.  I was realizing that as well, and thought a bit about alternatives, but didn't immediately think of a great one.  Can put some more thought into it.  Was considering copying some things into the component crate itself, but didn't think maintaining two copies of the code would be worth the tradeoff.

components/protocol-wasm/src/lib.rs
macovedj: Went ahead and did this, happy to rename if you think of something better

components/protocol-wasm/src/lib.rs
macovedj: Moved up to just after all the use statements, could probably move it to the line 1 too if you like

components/protocol-wasm/world.wit
macovedj: done

crates/server/src/api/proof.rs
macovedj: Nice, I've used it before, but haven't formed a habit of it yet.  Will definitely play with it some more though.  Deleted the `println!`.

components/protocol-wasm/src/lib.rs
macovedj: Oh yeah sorry, relic of just getting things to work.  I renamed to `ProtoBody`

components/protocol-wasm/world.wit
```diff
@@ -0,0 +1,87 @@
...
+    record map-checkpoint {
+      log-root: string,
+      log-length: u32,
+      map-root: string
+    }
+    record inclusion {
+      log: string,
+      map: string
+    }
+    validate: func(package-records: list<proto-envelope-body>) -> package-info
```
Kylebrown9, line 85: Turn validator into a "psuedo-resource" that can be initialized empty or with JSON starting contents and has methods that let you move the validator forward one package record at a time.

See wasi-streams [`input-stream`](https://github.com/WebAssembly/wasi-io/blob/main/wit/streams.wit#L30) for an example of a pseudo-resource.

components/protocol-wasm/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    match permission {
+        &package::Permission::Release => protocolbindings::Permission::Release,
+        &package::Permission::Yank => protocolbindings::Permission::Yank,
+        &_ => protocolbindings::Permission::Release,
+    }
+}
+
+impl protocolbindings::Protocol for Component {
+    fn prove_inclusion(input: protocolbindings::Inclusion, checkpoint: protocolbindings::MapCheckpoint, heads: Vec<protocolbindings::LogLeaf>) {
+      let map_checkpoint = MapCheckpoint {
```
Kylebrown9, line 82: Dead code, value `map_checkpoint` is not used.

macovedj: deleted

calvinrp:
Abandoning this PR. Very outdated. We have another contender that is forthcoming.

PR #95, opened 2023-04-19
Add configurable data stores to the server.

peterhuene:
This (large) PR implements a new `--data-store` option to the server.

The option defaults to `memory`, which stores logs and checkpoints in memory which will be lost when the server shuts down.

With the `postgres` feature enabled, it also supports a `postgres` value which requires the `DATABASE_URL` environment variable to point at the PostgreSQL server to use.

The `postgres` data store supports restarting the server and recreating the necessary data structures for the transparency service to work.

See `crates/server/README.md` for information on how to setup a persistent PostgreSQL store.


peterhuene:
@Kylebrown9 want to sync up tomorrow to do a high-level review walk-through before you dive into it?

crates/server/src/datastore/memory.rs
```diff
@@ -0,0 +1,491 @@
...
+                            sources: Default::default(),
+                        })
+                    }
+                    Err(e) => {
+                        log.validator.rollback(snapshot);
+                        Err(e.into())
+                    }
+                }
+            }
+            _ => Err(DataStoreError::RecordNotFound(record_id.clone())),
```
Kylebrown9, line 190: `DataStoreError::RecordNotFound` seems to be returned for records that are already accepted or rejected.

crates/server/src/datastore.rs
```diff
@@ -0,0 +1,211 @@
...
+    async fn initial_leaves(
+        &self,
+    ) -> Result<
+        Pin<Box<dyn Stream<Item = Result<InitialLeaf, DataStoreError>> + Send>>,
+        DataStoreError,
+    >;
+
+    /// Stores the given operator record.
+    ///
+    /// Returns the record id of the new record.
```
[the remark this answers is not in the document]
peterhuene: Indeed, a few of these comments are out of date as the trait evolved.  I'll clean up.

crates/server/src/datastore.rs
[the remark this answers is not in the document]
peterhuene: I'll move into `mod.rs`.

crates/server/src/datastore.rs
```diff
@@ -0,0 +1,211 @@
...
+    async fn reject_operator_record(
+        &self,
+        log_id: &LogId,
+        record_id: &RecordId,
+        reason: &str,
+    ) -> Result<(), DataStoreError>;
+
+    /// Accepts the given operator record.
+    ///
+    /// If the record fails to validate, it will be rejected.
```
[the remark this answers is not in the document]
peterhuene: That's correct; alternatively, I'm happy to have the stores return the validators so they can be externally validated and given back to the store to save when the record is updated.

crates/server/src/datastore.rs
```diff
@@ -0,0 +1,211 @@
...
+    async fn accept_operator_record(
+        &self,
+        log_id: &LogId,
+        record_id: &RecordId,
+    ) -> Result<(), DataStoreError>;
+
+    /// Stores the given package record.
+    ///
+    /// Returns the log leaf representing the new log record.
```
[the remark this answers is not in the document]
peterhuene: Indeed, another stale comment. I'll remove.

crates/server/src/datastore/memory.rs
```diff
@@ -0,0 +1,491 @@
...
+        .count()
+}
+
+/// Represents an in-memory data store.
+///
+/// Data is not persisted between restarts of the server.
+///
+/// Note: this is mainly used for testing, so it is not very efficient as
+/// it shares a single lock for all operations.
+pub struct MemoryDataStore(Arc<Mutex<State>>);
```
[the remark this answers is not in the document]
peterhuene: Couldn't hurt to change it to a `RwLock`.

crates/api/src/fetch.rs
```diff
@@ -17,8 +17,9 @@ pub struct FetchRequest {
     /// The root of the registry.
     pub root: DynHash,
     /// The last known operator record.
-    pub operator: Option<RecordId>,
+    pub since: Option<RecordId>,
```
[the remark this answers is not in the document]
peterhuene: I'll change the name back.

crates/api/src/fetch.rs
```diff
@@ -45,53 +47,44 @@ pub struct CheckpointResponse {
...
+    #[error("checkpoint `{checkpoint}` was not found")]
     CheckpointNotFound {
         /// The missing checkpoint.
-        checkpoint: Hash<Sha256>,
+        checkpoint: DynHash,
+    },
+    /// The log was not found.
+    #[error("log `{log_id}` was not found")]
+    LogNotFound {
+        /// The missing operator log id.
```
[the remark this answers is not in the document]
peterhuene: Originally I had the `DataStore` trait not take a `LogId` for operator log operations and just have it derive whatever the operator log id should be. I originally thought that perhaps that might be too limiting in the future (i.e. if the log ids changed for some reason), so instead I made the operator and package logs have parity with respect to the `DataStore` trait.

If we're okay with the `DataStore` deriving it, I'm happy to switch back.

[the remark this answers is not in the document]
peterhuene: Oh I missed that the comment was incorrect, yeah it should not be mentioning operator log here.

crates/server/src/datastore/postgres/migrations/2023-04-11-013700_initial_db/up.sql
```diff
@@ -0,0 +1,66 @@
...
+-- Unified table for both package and operator logs.
+-- The nullable columns are null for operator logs.
+-- TODO: add full text search indexes for description and keywords.
+CREATE TABLE logs (
+  id SERIAL PRIMARY KEY,
+  log_id TEXT NOT NULL UNIQUE,
+  name TEXT NOT NULL, -- implied UNIQUE constraint as log_id is derived from name
+  validator JSONB NOT NULL,
+  description TEXT,
+  keywords TEXT,
```
Kylebrown9, line 25: remove for now

crates/server/src/datastore.rs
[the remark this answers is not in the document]
peterhuene: I renamed the "accepted" state to "validated", which I hope clears up that the data store is the one validating the records into the log.

crates/server/src/datastore/memory.rs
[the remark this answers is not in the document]
peterhuene: Changed to a `RwLock`.

crates/server/src/datastore/memory.rs
peterhuene: Added a `RecordNotPending` error variant for this case.

crates/server/src/datastore/postgres/migrations/2023-04-11-013700_initial_db/up.sql
peterhuene: The columns were removed.

crates/protocol/src/operator/validate.rs
```diff
@@ -113,31 +113,29 @@ impl Validator {
...
-    fn validate_envelope(
+    /// # Safety
+    ///
+    /// This is marked as unsafe as the caller must ensure that the validator
+    /// is snapshotted before calling this method and rolled back if the
+    /// validation fails.
+    ///
+    /// Failure to do so may result in the validator being left in an inconsistent
+    /// state.
+    pub unsafe fn validate_record(
```
Kylebrown9, line 138: IMO either this isn't actually `unsafe` or we need to change it. A basic API like this shouldn't be designed in a way that needs `unsafe`.

crates/protocol/src/serde_envelope.rs
```diff
@@ -13,6 +13,17 @@ pub struct SerdeEnvelope<Contents> {
 }
 
 impl<Contents> SerdeEnvelope<Contents> {
+    /// Creates a new envelope for some contents using a signature.
+    ///
+    /// Note: this does not verify the signature matches the contents.
+    pub fn new(contents: Contents, key_id: signing::KeyID, signature: signing::Signature) -> Self {
```
Kylebrown9, line 19: Is this needed somewhere? I would prefer if the only way to make these guaranteed signature consistency.

crates/server/src/api/package.rs
```diff
@@ -233,14 +212,20 @@ async fn get_pending_record(
...
-        &record_id,
-        status,
-    )?))
+    Ok(Json(
+        match config
+            .core_service
+            .get_record_status(&package_id, &record_id)
+            .await?
+        {
+            // Validated is considered still pending until the record is part of a checkpoint
```
Kylebrown9, line 221: Should read "Validated is considered still processing"

crates/server/src/datastore/postgres/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+    operator, package,
+    registry::{LogId, LogLeaf, MapCheckpoint, RecordId},
+    ProtoEnvelope, SerdeEnvelope,
+};
+
+mod models;
+mod schema;
+
+/// Trait for abstracting operator/package validator interactions for persistence.
+trait Validator: std::fmt::Debug + Serialize + DeserializeOwned + Default + Send + Sync {
```
Kylebrown9, line 33: Should this be defined in the `warg-protocol` crate?

crates/server/src/datastore/postgres/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+                })?;
+
+            for source in sources {
+                let (kind, url) = match &source.kind {
+                    ContentSourceKind::HttpAnonymous { url } => {
+                        (SourceKind::Http, Some(url.as_str()))
+                    }
+                };
+
+                diesel::insert_into(schema::sources::table)
```
Kylebrown9, line 205: Is there not an insert-many?

crates/server/src/services/transparency/log.rs
```diff
@@ -24,39 +26,50 @@ pub struct Summary {
     pub log_length: u32,
 }
 
-pub fn process(input: Input) -> Output {
-    let (summary_tx, summary_rx) = mpsc::channel::<Summary>(4);
-    let (log_data_tx, log_data_rx) = mpsc::channel::<LogLeaf>(4);
+pub fn spawn(input: Input) -> Output {
```
Kylebrown9, line 29: I think the original distinction is that since this spawns a new task on (old 31/new 33) and doesn't do all the work itself we were calling it spawn. What sort of distinction are you following?

tests/postgres/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    let (_server, config) = super::support::spawn_server(root, Some(data_store()?)).await?;
+    callback(config).await
+}
+
+/// This test assumes the database is empty on each run.
+/// Use the `ci/run-postgres-tests.sh` script to run this test.
+///
+/// A smoke test that ensures that postgres integration works.
+///
+/// This is one large test currently because there can be only one core service
```
Kylebrown9, line 28: Is there a reason we can't run two completely independent servers in parallel?

crates/protocol/src/serde_envelope.rs
peterhuene: We don't guarantee the signature on deserialization either; this is just a way of constructing an envelope from the constituent parts without using serialization.

tests/postgres/mod.rs
peterhuene: Each server would independently have their own log and map cryptographic primitives to checkpoint from; thus the transparency services of each running server would have disjoint leafs (consequently, different log lengths and roots).

If you don't anticipate a problem with that, then I don't see a reason why not then.

If that would be a problem and we do require a singular transparency service, then I think we'll need to refactor the core service out to another process and use a proper (and reliable) message broker between them.

crates/protocol/src/serde_envelope.rs
Kylebrown9: IMO, not guaranteeing on serialization/de-serialization is different because as long as you're not synthesizing serialized versions without using the serde it's still always going to be valid. This makes it possible to mess up in a more typical usage. As a compromise, can we make this `pub(crate)` or restricted in some other way?

crates/server/src/services/transparency/log.rs
peterhuene: Yeah, that's basically the reasoning behind the change; this function spawns a new task from the input, giving it the output of the spawned task (i.e. the task handle, senders, etc.).

The name `process` to me implies it's doing the work itself; i.e. something that should be called repeatedly to "do the thing".

tests/postgres/mod.rs
Kylebrown9: We do need a single transparency service per registry instance, but I wasn't sure what this comment was getting at. It doesn't seem to me like there's anything keeping you from splitting this test into multiple tests that initialize their own completely independent registry instances and then each test one thing (like publish/validate). It does mean that you'd spin up several whole registries and their own databases, but if that's the problem I would write the comment differently.

crates/server/src/services/transparency/log.rs
Kylebrown9: Oh, my bad. I read the diff backwards. LGTM

tests/postgres/mod.rs
peterhuene: Oh gotcha! Yeah, the comment is in the context that the tests are assuming there's a single _database_ being communicated with for all of the tests, as the CI script just spawns a single container. Thus while we can start multiple servers to test against concurrently, the data would be shared across them with inconsistent core service state.

I can update the comment to make it less confusing.

crates/protocol/src/serde_envelope.rs
peterhuene: Agreed that having a `new` method vs. deserializing does make it more of a footgun.

It can't be `pub(crate)` as it's called from the server crate, but perhaps I can figure out a different approach.

Taking a step back, though, perhaps it's not valuable to have the individual columns on the `checkpoints` table and instead just persist the envelope itself as a JSON blob (like we do for validator state)?

Let me see what I can come up with.a

crates/protocol/src/operator/validate.rs
```diff
@@ -11,52 +11,52 @@ use warg_crypto::{signing, Signable};
 
 #[derive(Error, Debug)]
 pub enum ValidationError {
-    #[error("The first entry of the log is not \"init\"")]
+    #[error("the first entry of the log is not \"init\"")]
```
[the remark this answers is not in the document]
peterhuene: There's no definitive guideline I am aware of, but generally errors like these tend to be sentence fragments and may end up being formatted with colons all on a single line (e.g. "I was doing X but encountered an error: {e}").

That'd be my only pitch for why they should be lowercase; I'm not of a strong opinion, although I personally prefer it.

crates/server/src/datastore/postgres/mod.rs
peterhuene: Happy to move it there.

crates/server/src/datastore/postgres/mod.rs
peterhuene: Apparently `values` can take an iterator for multiple value inserts, so I'll fix this.

crates/protocol/src/operator/validate.rs
peterhuene: I think I can remove this method; it was introduced early on with these changes as a way to prevent repeated validator snapshotting, but the remaining place we do an explicit `snapshot` call isn't necessary any more.

crates/protocol/src/serde_envelope.rs
peterhuene: I couldn't come up with something that wasn't horrible, and I do like that we have the columns in the `checkpoints` table rather than some serialized blob for the envelope, so I renamed the method to `from_parts_unchecked` to make it a little clearer that this method shouldn't be used to construct any ol' `SerdeEnvelope`.

Thoughts on a different approach?

crates/protocol/src/operator/validate.rs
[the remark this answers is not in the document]
Kylebrown9: that works for me

crates/protocol/src/serde_envelope.rs
Kylebrown9: I think that's a good compromise. I like that `unchecked` makes it clear you can do this wrong if you aren't checking things yourself.

Kylebrown9 (approved):
Thanks for putting this together!

PR #96, opened 2023-04-26
Create initial containerization

### LINKED ISSUE #46: Task: create container images (and docker compose?) to easily stand-up a local server for development

kriswuollett:
- created Dockerfile and Docker Copmose files to cover server, client, and potential utilities like database migration package
- added scripts in `infra/local` to act as both documentation and ease of use for trying out and developing the registry locally
- renamed/updated the operator key environment variable so that it would more usable in a cloud environment, e.g., the private key stored in a `k8s` `Secret` which is mounted in a volume
- resolves #46

CleanShot 2023-04-27 at 09 11 48@2x

CleanShot 2023-05-03 at 16 09 13@2x

CleanShot 2023-05-03 at 16 21 45@2x


local-infra.sh
```diff
@@ -0,0 +1,45 @@
...
+CMD="${1:-}"
+if [[ "$CMD" == "" ]]; then
+  printf "Must specify command, either 'up', 'down', or 'rm'\n" 1>&2
+  exit 1
+fi
+
+case "$CMD" in
+up)
+  docker-compose -f "$REPO_DIR/docker-compose.yaml" -f "$REPO_DIR/docker-compose.postgres.yaml" up --build -d
+  PGPORT=$(docker port bytecodealliance-registry-db-1 5432 | sed -e 's/.*://g')
```
kriswuollett, line 17: I think it may be better to just set a non-5432 port to avoid conflicts, and just suggest to use an additional docker-compose override file to change the port if necessary. That could remove the complication of getting the random port here. Although keeping the random value generation for passwords/keys seems like more appropriate practice to follow.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+WORKDIR /usr/src/bytecodealliance/registry
+COPY --from=planner /usr/src/bytecodealliance/registry/recipe.json ./
+RUN cargo chef cook --release --workspace --recipe-path recipe.json
+COPY . .
+RUN cargo build --release --workspace
+
+FROM gcr.io/distroless/cc AS warg-server
```
kriswuollett, line 25: It is probably seen more often to have a separate Dockerfile for a separate binary, but having separate targets here are maybe helping speed and simplify integration tests?

docker-compose.yaml
```diff
@@ -0,0 +1,38 @@
...
+        fi
+    volumes:
+      - warg-secrets:/var/secrets/warg:rw
+  warg-registry:
+    build:
+      target: warg-server
+    ports:
+      - "8090"
+    environment:
+      - WARG_OPERATOR_KEY_FILE=/var/secrets/warg/operator-key
```
kriswuollett, line 23: In `k8s` this would exist as the value of key `operator-key` in a `Secret` mounted at `/var/secrets/warg`. It's probably good to check the different ways of configuring some options works.

kriswuollett (commented):
Self-review comments, perhaps some need to be code comments?

local-infra.sh
kriswuollett: Created `.env` for defining default port numbers.

crates/server/src/main.rs
```diff
@@ -38,8 +43,29 @@ async fn main() -> Result<()> {
     args.init_tracing();
     tracing::debug!("Args: {args:?}");
 
-    let demo_operator_key = std::env::var("WARG_DEMO_OPERATOR_KEY")?;
-    let signing_key = demo_operator_key.parse()?;
+    // TODO(#46): Add --warg-demo-operator-key[-file] and others cli options
```
Kylebrown9, line 46: This looks like it could be its own function called e.g. `get_operator_key`

docker-compose.postgres.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - -c
+      - >- 
+        cd /var/secrets/warg &&
+        if [[ ! -f postgres-admin-password ]]; then
+          < /dev/urandom tr -dc '[:alnum:]' | head -c16 >postgres-admin-password
+        fi
+    volumes:
+      - warg-secrets:/var/secrets/warg:rw
+  db:
+    # NOTE: Note as of 2023-04-26 not all IaaS supports v15 for managed SQL
```
Kylebrown9, line 16: (Non-blocking) Any reason in particular?

Kylebrown9 (commented):
Looks great, thanks for helping set this all up. The utilities around set up tear down and connecting to the DB look pretty convenient. I won't approve it quite yet so that we can get one more pair of eyes on it before it goes in since I don't do much bash.

@lann can you take a quick look over this?

Dockerfile
Kylebrown9: @peterhuene has done the most integration testing work, so I'll let him speak to this.

crates/server/src/main.rs
kriswuollett: Yes, I expected something like this would be needed soon. I'm not familiar with what's available in Rust libraries in general yet. I'd expect it perhaps be a bit more generic where selected props come from either a command line argument or an environment variable in that order, with preference given to a "file" variation if set for a limited set of things like passwords, etc.

docker-compose.postgres.yaml
kriswuollett: The secrets volume, or the database version? I was just being pedantic about not having an easy to guess password for something that may be opening up a port on one's machine and showing how to not pass it as an environment variable, and also assumed that the project isn't concerned about running on a specific version range of PostgreSQL yet. Just wanted to make sure it would be compatible for anyone running on in the cloud, e.g., [DO](https://docs.digitalocean.com/products/databases/postgresql/details/limits/) (latest v14), [GCP](https://cloud.google.com/sql/docs/postgres/db-versions) (latest v14), [AWS](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html) (latest v15), [Azure](https://learn.microsoft.com/en-us/azure/postgresql/single-server/concepts-version-policy#supported--postgresql-versions) (latest v14), etc.

Dockerfile
lann, line 25: For debugging purposes it can be quite helpful to have a shell and debug tools available on server containers. Is there another way to do that with distroless images?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /usr/src/bytecodealliance/registry
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+WORKDIR /usr/src/bytecodealliance/registry
+COPY --from=planner /usr/src/bytecodealliance/registry/recipe.json ./
+RUN cargo chef cook --release --workspace --recipe-path recipe.json
+COPY . .
+RUN cargo build --release --workspace
```
lann, line 23: Curious why these two stages are split?

lann (approved):
Thanks for the contribution! I don't think we need to worry too much about the specifics of this until we're getting closer to running production; I'm sure there will be more feedback once people actually use this code.

Dockerfile
kriswuollett: Yes, I'd suggest we could add an `ARG` to do a [debug variant](https://github.com/GoogleContainerTools/distroless#debug-images) which is in turn used in the `docker-compose`, but otherwise the `Dockerfile` would build a release version.

kriswuollett: Probably a separate target would be better, that we can also do something like make it remotely debuggable, e.g., debug build and include `gdbserver` in the container and use it for the entrypoint? Follow up PR I'd say...

Kylebrown9:
Let's hold merging this till Peter's PR is in. He has a lot of changes and it'll be easier for this PR to rebase on to his than the other way round.

kriswuollett:
Cool, sounds good, I can take care replacing the database migration stand-in then.

Kylebrown9:
Sounds good. #95 just went in, so that should be possible now.

kriswuollett:
OK, will take a look to update this PR.

kriswuollett:
OK, think should be good now. Later PRs I can do merge commits if required, but I do like re-based minimal specific commits with no extras. :-)

kriswuollett:
@Kylebrown9 incorporated the new `postgres` data store into container setup, as well as added a `warg-server-debug` target that can be substituted if needed that contains debugging tools. Also switched to using `debian:bullseye-slim` to include `libpq` for the server.

Dockerfile
Kylebrown9, line 1: Add a comment to each stage describing its purpose.

crates/server/src/bin/warg-server.rs
```diff
@@ -52,7 +58,27 @@ async fn main() -> Result<()> {
     tracing::debug!("args: {args:?}");
 
     // TODO: pull the signing key from the system keyring
-    let operator_key = std::env::var("WARG_DEMO_OPERATOR_KEY")?.parse()?;
+    // TODO: Add --warg-operator-key[-file] and others cli options
+    let operator_key: PrivateKey = match std::env::var("WARG_OPERATOR_KEY_FILE") {
```
Kylebrown9, line 62: Extract into a function named e.g.  `get_operator_key`.

infra/local/locals.inc.sh
```diff
@@ -0,0 +1,31 @@
...
+  SCRIPT_DIR="$1"
+  if [[ ! -d "$SCRIPT_DIR" ]]; then
+    printf "generate_locals missing arguments (script_dir)\n" 1>&2
+    return 1
+  fi
+
+  PGPASS_FILE="$SCRIPT_DIR/pgpass.local.conf"
+  PSQL_FILE="$SCRIPT_DIR/psql.local.sh"
+
+  PGPORT=$(docker port bytecodealliance-registry-db-1 5432 | sed -e 's/.*://g')
```
Kylebrown9, line 15: A. Did this change in a recent commit? I thought `PGPASS` was randomized.
B. Can you add a comment that documents what this section does, links any references for the pattern, and comments on the security?

kriswuollett:
Working on changes, will reply when finish adding patch for items above.

Dockerfile
kriswuollett: I created a `warg-server-debug` Dockerfile target for convenience if a developer wants have more tools on the vm than just `bash` and other Linux basics.

crates/server/src/bin/warg-server.rs
kriswuollett: I extracted a function `get_operator_key` from which I also extracted the arg (file)/env processing behavior into a function `get_opt_content` in a crate `warg-args-util` which can also be picked up by the `warg` CLI client or other apps if desired. Used `get_opt_content` to also pick up the database URL from a file since it contains sensitive data.

infra/local/locals.inc.sh
kriswuollett: A. Yes, I backtracked on reading sensitive information from files since I was thinking of doing it in a follow up PR. Instead I've just done it as properly as I think I can at the moment with the latest additional patches.
B. In the latest patches I added and/or fixed the doc comment at the top of the file.

kriswuollett:
OK, ready for review again.

crates/server/src/bin/warg-server.rs
Kylebrown9: I think splitting it out into its own entire crate is premature. We should probably wait to do that in a PR wherein we actually want to call this function from somewhere else.

kriswuollett: Added patch to undo the new library creation.

Dockerfile
Kylebrown9: Last blocker for approve & merge

kriswuollett: Added Dockerfile target docs.

Kylebrown9:
Thanks! This looks great. I'll send it through once the CI finishes.

Kylebrown9 (approved):

PR #98, opened 2023-04-29
Protocol readily usable by other platforms

### LINKED ISSUE #97: `warg.proto` not in appropriate directory path
The [package](https://protobuf.dev/programming-guides/proto3/#packages) is `warg`:

https://github.com/bytecodealliance/registry/blob/eee527f91d96776b58cf4631911eecabdee7512e/crates/protocol/src/warg.proto#L1-L5

Therefore the `warg.proto` file should be in a folder named `warg` that would be [included in a `protoc -I` path](https://protobuf.dev/programming-guides/proto3/#generating). Whether or not package `warg` is appropriate would be another issue, but it does seem unique enough and probably would be the C API import path.

Suggesting moving `src/warg.proto` to `src/warg/warg.proto` so non-Rust `protoc` usages can properly resolve it.

kriswuollett:
- resolves #97


crates/protocol/build.rs
```diff
@@ -1,7 +1,7 @@
 use std::{env, io::Result, path::PathBuf};
 
 fn main() -> Result<()> {
-    let warg_proto = PathBuf::from("src/warg.proto");
+    let warg_proto = PathBuf::from("src/warg/warg.proto");
```
peterhuene, line 4: We should probably take this opportunity to move it to a top-level directory rather than under `src/`, as that's where I expect Rust source to live, not protobuf definitions.

Perhaps `crates/protocol/warg/warg.proto`?

kriswuollett: Cool, yeah that seems good.

Kylebrown9: I actually want to take this a step further. If the point of this is to make the `.proto` accessible outside the Rust crate to other consumers, then it probably shouldn't be in the `crates` directory at all. We should probably put this file in some root folder with the spec when it's written.

kriswuollett: That's great -- I did want to try proposing how it could be more cross-platform usable -- so I'll just drop the minor fix steps that were just nice to haves, and make a PR that would have all steps to discuss.

Kylebrown9:
As discussed in a comment above, this PR will move the protos outside the `crates` directory to a new top-level directory structure for the protocol spec and proto files. This will make it more usable for other implementations.

The current structure of the directory is
```
+ protocol
   README.md
   + protos
      + google/protobuf
         timestamp.proto
      + warg
         warg.proto
```

@lann @peterhuene any opinions about the name/structure of this new top-level directory?

kriswuollett:
I'd probably have preferred `api` as the top level directory, but kept it the same as the existing crate `protocol` since there already was an `api` crate. Also not sure if you'd prefer `proto` or `protos`?

peterhuene:
@Kylebrown9 the proposed structure works for me.

Kylebrown9:
I'm happy to move forward with Peter and I approving the structure.

Feel free to remove the draft status when you're ready and we can do a final review.

kriswuollett:
Still reading through the code - the [proofs](https://github.com/bytecodealliance/registry/blob/78b46a810d3db52706863f7e3f9493daad70e4b7/crates/transparency/src/proofs.proto) protos applies as well if they are serialized somewhere in there I did not see yet? Is having them in a separate namespace needed?

Kylebrown9:
No, the separate `proofs` namespace shouldn't have any functional importance.

lann:
I think the most common convention is a top level `proto/` directory, but as long as everything is package-namespaced under _some_ root I don't think it makes much difference functionally.

Kylebrown9:
On the subject of `proofs` one of the reason it's separate is that we only need types generated for `proofs` in the `warg-transparency` crate and `warg` in the `warg-protocol` crate. Any change that moves `proofs` will need to preserve that.

kriswuollett:
Ready for review again. Also moved the transparency proofs protos but kept them in separate package from the protocol protos.

proto/google/protobuf/timestamp.proto
Kylebrown9, line 1: Is it typical to bring proto dependencies into the source tree of your own repo?

Kylebrown9 (approved):

PR #100, opened 2023-04-30
Use gRPC to define protocol and REST mapping with OpenAPI spec

### LINKED ISSUE #99: Recommend API routes have `v1` prefix
Some guides [recommend that APIs are versioned](https://cloud.google.com/apis/design/versioning) so that breaking changes can be introduced later. Recommend that start with a `v1` prefix. Currently there is no prefix:

https://github.com/bytecodealliance/registry/blob/eee527f91d96776b58cf4631911eecabdee7512e/crates/server/src/lib.rs#L234-L237

kriswuollett:
Just a draft, but perhaps good for discussion on API scoping since the protocol can roughly be seen in a few non-Rust files in `protocol/protos` -- it includes REST mapping with the JSON encoding as [officially defined](https://protobuf.dev/programming-guides/proto3/#json) (just how current usage of `pbjson` also does). 

- stacked upon #98
- resolves #99 
- creates a definition of `warg` protocol into single `proto` file
- uses [`tonic`](https://docs.rs/tonic/0.9.2/tonic) gRPC server with `grpc`-named Rust feature providing a REST interface via [`tonic_web`](https://docs.rs/tonic-web/0.9.2/tonic_web) deprecating `axios` usage
- uses [`grpc-gateway`](https://github.com/grpc-ecosystem/grpc-gateway) `protoc` plugin to generate REST interface spec in OpenAPI format
- adds [gRPC health checking protocol](https://github.com/grpc/grpc/blob/b1bdf03d406ce256401564674c5b240f280af9ff/doc/health-checking.md) to server on (PR TBD) existing REST health check URLs


peterhuene:
Hi @kriswuollett.

Could you expand on the motivation for these changes a little more? We don't plan on implementing support for gRPC in this implementation of the registry, so I'm not entirely sure why we would want to derive the REST API spec from a gRPC-based definition.

The original intention of using protobuf in this design was scoped to using it only for an extensible log record serialization format.

That said, I am also currently drafting an OpenAPI spec so that we can come to a final specification of the REST API so that progress can be made on compatible, alternative implementations.

Kylebrown9:
There has been some interest in having a gRPC API, but it's not high on the priority list and I think people are going to, reasonably, be resistant to trying to derive the REST API from it given that the proto-to-JSON mapping makes weird JSON.

API design, specification, and versioning is going to be a topic for tomorrows SIG-Registries meeting, so we can chat more about it there and then continue here.

kriswuollett:
I was seeing if the REST and OpenAPI spec could be derived from gRPC automatically in Rust, mostly as a learning exercise for myself if I were to build other servers too in Rust, but if it worked it may be a time saver for the project.

Unfortunately I'm now seeing it is not possible for automatic mapping due to the following which I don't think will get resolved anytime soon.

- https://github.com/tokio-rs/prost/issues/592
- https://github.com/hyperium/tonic/issues/758

From the issues it looks like the `http` annotations may not be used by the `tonic-web` mapping, therefore generating an [OpenAPI spec from the Go grpc-gateway project won't work](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/) won't match up.

So I'll probably be closing this pull request soon.

kriswuollett:
I haven't used `hyper` to have an opinion on it, but this PR shows a proof-of-concept from `.proto` to `service.swagger.json` (OpenAPI) to generated `hyper` client and server stubs. Keeps gRPC/REST client/server stubs in sync with a spec, rather than hand-writing code to match a spec.

Although I'd consider a server easier to maintain if it was gRPC only and just use build-in translation or Go's `grpc-gateway` as a proxy if the languages gRPC-Web isn't compliant with the spec.

peterhuene:
For reference, here's a [link](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/bytecodealliance/registry/721de674fe471774ad1ca1bcc1e75cdb6313568b/crates/rest_listener/api/openapi.yaml) to visualize the OpenAPI spec that was generated from the protos.

kriswuollett:
Cool, the OpenAPI visualization page is great.

Even though it wouldn't be customizable, what were the things that aren't quite standard practice for the [JSON mapping](https://protobuf.dev/programming-guides/proto3/#json)? Is it just the all-caps snake case for the enum values? Or [oneof](https://github.com/protocolbuffers/protobuf/issues/3928) being more like a union?

Kylebrown9:
@kriswuollett can you generate OpenAPI 3.0 for your version and use [ReDoc](https://redocly.github.io/redoc/) to visualize it so we can do a comparison?

kriswuollett:
There seems to be another one generated at `crates/rest_listener/api/openapi.yaml`. But I also will check out https://github.com/google/gnostic for doing the reverse OpenAPI v3 to gRPC protos when I get a chance.

calvinrp:
Let's hold on the gRPC implementation for now. Revisit later. Focus on completing the REST API.

PR #101, opened 2023-05-03
Add an `openapi.yaml` file to the server implementation.

peterhuene:
[Rendered API documentation](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/bytecodealliance/registry/62d8a9cfaa4ce424740c10d76d640b906836d753/crates/server/openapi.yaml)

This PR creates a single OpenAPI specification for the Warg registry server.

It is intended to (ultimately) be the complete specification of a Warg registry server's REST API.

The specification proposed here is not what is currently implemented by `warg-server`, but it is inspired heavily by both the existing implementation and previously proposed APIs.

We can use this as a starting point to finalize an API specification for the initial version.

An API around streaming records to federated registries is not currently included in the specification, so design around this is welcome.

### DIFF
--- a/crates/server/openapi.yaml
+++ b/crates/server/openapi.yaml
@@ -687,8 +687,14 @@ components:
       description: A request to append a record to a package log.
       additionalProperties: false
       required:
+        - name
         - record
       properties:
+        name:
+          type: string
+          description: The name of the package log being appended to.
+          maxLength: 128
+          example: wasi:http
         record:
           "$ref": "#/components/schemas/EnvelopeBody"
           description: The package record being appended to the log.
--- a/crates/server/openapi.yaml
+++ b/crates/server/openapi.yaml
@@ -201,7 +201,7 @@ paths:
           content:
             application/json:
               schema:
-                "$ref": "#/components/schemas/PublishPackageRecordResponse"
+                "$ref": "#/components/schemas/PackageRecord"
         default:
           description: An error occurred when processing the request.
           content:
@@ -508,7 +508,6 @@ paths:
           content:
             application/json:
               schema:
-                type: object
                 oneOf:
                   - "$ref": "#/components/schemas/BundleFailureError"
                 discriminator:
@@ -578,7 +577,6 @@ paths:
           content:
             application/json:
               schema:
-                type: object
                 oneOf:
                   - "$ref": "#/components/schemas/PackageNotIncludedError"
                   - "$ref": "#/components/schemas/IncorrectProofError"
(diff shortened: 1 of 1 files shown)


peterhuene:
The intention behind this PR is to be a starting point for coming to consensus on exactly what the REST API should be for 1.0.

I'll bring this up in tomorrow's meeting, but everyone should feel empowered to collaborate on this proposed specification.

The sooner we have a real spec, the sooner we can get the `warg-server` implementation conforming and alternative implementations can begin.

Kylebrown9:
There are things here I think make a lot of sense, like putting limits on package log fetching which works pretty well with the "since"-semantics to enable paging/cursoring.

There are other changes here that I'm less sure about and I think are going to require a lot of discussion. Especially around changes to the package API and the introduction of metadata.

In the package log semantics, you can't initialize a package without signing a record containing an init entry and there's nothing preventing you from putting other entries in that record. So IMO it doesn't really make sense to separate initialization into its own API and if you did that API would still need to send a signed record.

The package metadata proposed here isn't verifiable by end users and its not clear who would have permission to update it. We should discuss different options for metadata, because there are ways to include metadata that are more consistent with the rest of the protocol design and in-log permissions system.

Other than that, there's some minutiae. For example, I think the `name` field in the 200 response to `GET /package/{logId}` might need to be optional someday in the case that we eventually support name-takedown and I'm not sold on `appended` as the name of that status, but we can bikeshed these after the core issues are resolved.

peterhuene:
I'm totally fine with eliminating the "create package" and "update package" endpoints in favor of putting the metadata into the log itself, thus only needing a "append record" endpoint for both creating new logs and updating package metadata. I'd like us to start thinking about basic metadata sooner than later, though.

We might still want a "give me the latest metadata" endpoint (i.e. `GET /package/{logId}`) where the registry can return back what it believes to be the latest metadata; clients could verify it is correct themselves via validating the log, if so desired.

I'm definitely not married to `appended`, so open to alternatives; `accepted` I wanted to avoid because it can be confused with the HTTP status and `published` didn't feel right for records when most registries talk about releases in that context.

crates/server/openapi.yaml
```diff
@@ -1026,6 +1015,21 @@ components:
           minLength: 107
           maxLength: 107
           example: "ecdsa-p256:MEUCIQCzWZBW6ux9LecP66Y+hjmLZTP/hZVz7puzlPTXcRT2wwIgQZO7nxP0nugtw18MwHZ26ROFWcJmgCtKOguK031Y1D0="
+    ContentSourceMap:
+      type: object
+      description: The map of content digest to sources.
+      patternProperties:
+        "^sha256:[a-f0-9]{64,64}$":
```
Kylebrown9, line 1022: This pattern would have to change in future versions that support other algorithms. Is there a better way of generalizing this / do we need to encode the algorithm constraint here at all?

peterhuene: Yeah, I think I'll just relax these patterns to be more generic, something like `[a-z0-9-]+:[a-f0-9]+` for `AnyHash` and change signatures similarly (but keep the base64 regex after the colon).

The API spec should be algorithm agnostic.

peterhuene:
With #114 merged, this spec now reflects the current state of the REST API implemented in the client and server.

peterhuene:
It's up to date in the sense it describes how the current implementation works.

I think we should merge it to then let Danny update it with the proposed changes to return the upload URLs to the client.

calvinrp (approved):

Issue #82, opened 2023-03-22
Check Log-Checkpoint Consistency


Kylebrown9:
When the client updates, use the consistency proof API to check that the new checkpoint is consistent with the old one.

peterhuene:
This was implemented in #102.

Issue #83, opened 2023-03-22
Client Must Read Operator Log


Kylebrown9:
The client must read the operator log and use the resulting key to verify the checkpoint.

macovedj:
May be time to close this?

PR #102, opened 2023-05-03
client consistency proof

macovedj:
Addresses https://github.com/bytecodealliance/registry/issues/82

### DIFF
--- a/crates/api/src/proof.rs
+++ b/crates/api/src/proof.rs
@@ -91,6 +91,15 @@ pub enum ProofError {
         /// The id of the package.
         id: LogId,
     },
+
+    /// Failed to prove consistency of a package.
+    #[error("failed to prove consistency of `{old_root}` with `{new_root}`")]
+    LogNotConsistent {
+        /// Old Root
+        old_root: DynHash,
+        /// New Root
+        new_root: DynHash,
+    },
     /// The provided root for an inclusion proof was incorrect.
     #[error("failed to prove inclusion: found root `{found}` but was given root `{root}`")]
     IncorrectProof {
--- a/crates/api/src/proof.rs
+++ b/crates/api/src/proof.rs
@@ -92,7 +92,7 @@ pub enum ProofError {
         id: LogId,
     },
 
-    /// Failed to prove consistency of a package.
+    /// Failed to prove consistency of the registry between two checkpoints.
     #[error("failed to prove consistency of `{old_root}` with `{new_root}`")]
     LogNotConsistent {
         /// Old Root
--- a/crates/client/src/api.rs
+++ b/crates/client/src/api.rs
@@ -22,7 +22,7 @@ use warg_protocol::{
     ProtoEnvelopeBody,
 };
 use warg_transparency::{
-    log::{LogData, LogProofBundle, ProofBundle, VecLog},
+    log::{LogProofBundle, ProofBundle},
     map::MapProofBundle,
 };
 
--- a/crates/client/src/api.rs
+++ b/crates/client/src/api.rs
@@ -10,7 +10,10 @@ use warg_api::{
     content::{ContentError, ContentResult, ContentSource},
     fetch::{CheckpointResponse, FetchError, FetchRequest, FetchResponse, FetchResult},
     package::{PackageError, PackageResult, PendingRecordResponse, PublishRequest, RecordResponse},
-    proof::{InclusionRequest, InclusionResponse, ProofError, ProofResult},
+    proof::{
+        ConsistencyRequest, ConsistencyResponse, InclusionRequest, InclusionResponse, ProofError,
+        ProofResult,
+    },
     FromError,
 };
 use warg_crypto::hash::{DynHash, Sha256};
@@ -219,9 +222,19 @@ impl Client {
         old_root: DynHash,
         new_root: DynHash,
     ) -> ProofResult<()> {
-        dbg!(old_root);
-        dbg!(new_root);
-        todo!()
+        let request = ConsistencyRequest { old_root, new_root };
+        let url = self.url.join("proof/consistency").unwrap();
+        let response = into_result::<ConsistencyResponse, ProofError>(
+            self.client
+                .post(url)
+                .json(&request)
+                .send()
+                .await
+                .map_err(ProofError::from_error)?,
+        )
+        .await?;
+        let proof = response.proof;
+        Ok(())
     }
 
     /// Uploads package content to the registry.
--- a/crates/client/src/api.rs
+++ b/crates/client/src/api.rs
@@ -21,7 +21,10 @@ use warg_protocol::{
     registry::{LogLeaf, MapCheckpoint, MapLeaf},
     ProtoEnvelopeBody,
 };
-use warg_transparency::{log::LogProofBundle, map::MapProofBundle};
+use warg_transparency::{
+    log::{LogData, LogProofBundle, ProofBundle, VecLog},
+    map::MapProofBundle,
+};
 
 async fn deserialize<T: DeserializeOwned>(response: Response) -> Result<T, String> {
     let status = response.status();
@@ -221,7 +224,11 @@ impl Client {
         &self,
         old_root: DynHash,
         new_root: DynHash,
+        old_length: u32,
+        new_length: u32,
     ) -> ProofResult<()> {
+        let old = old_root.clone();
+        let new = new_root.clone();
         let request = ConsistencyRequest { old_root, new_root };
         let url = self.url.join("proof/consistency").unwrap();
         let response = into_result::<ConsistencyResponse, ProofError>(
@@ -233,7 +240,13 @@ impl Client {
                 .map_err(ProofError::from_error)?,
         )
         .await?;
-        let proof = response.proof;
+        let proof = ProofBundle::<Sha256, [u8; 32]>::decode(&response.proof).unwrap();
+        let unbundled = proof.unbundle().0;
+        let consistency_proof =
+            unbundled.prove_consistency(old_length as usize, new_length as usize);
+        let (found_old, found_new) = consistency_proof.evaluate(&unbundled).unwrap();
+        assert_eq!(old, DynHash::from(found_old));
+        assert_eq!(new, DynHash::from(found_new));
         Ok(())
     }
 
--- a/crates/client/src/api.rs
+++ b/crates/client/src/api.rs
@@ -224,8 +224,6 @@ impl Client {
         &self,
         old_root: DynHash,
         new_root: DynHash,
-        old_length: u32,
-        new_length: u32,
     ) -> ProofResult<()> {
         let old = old_root.clone();
         let new = new_root.clone();
@@ -240,14 +238,35 @@ impl Client {
                 .map_err(ProofError::from_error)?,
         )
         .await?;
-        let proof = ProofBundle::<Sha256, [u8; 32]>::decode(&response.proof).unwrap();
-        let unbundled = proof.unbundle().0;
-        let consistency_proof =
-            unbundled.prove_consistency(old_length as usize, new_length as usize);
-        let (found_old, found_new) = consistency_proof.evaluate(&unbundled).unwrap();
-        assert_eq!(old, DynHash::from(found_old));
-        assert_eq!(new, DynHash::from(found_new));
-        Ok(())
+        let proof = ProofBundle::<Sha256, LogLeaf>::decode(&response.proof).unwrap();
+        let (log_data, consistencies, inclusions) = proof.unbundle();
+        if !inclusions.is_empty() {
+            return Err(ProofError::BundleFailure {
+                message: "Expected no inclusion proofs".into(),
+            });
+        }
+        if consistencies.len() != 1 {
+            return Err(ProofError::BundleFailure {
+                message: "Expected exactly one consistency proof".into(),
+            });
+        }
+        let (old_hash, new_hash) =
+            consistencies
+                .first()
+                .unwrap()
+                .evaluate(&log_data)
+                .map_err(|_| ProofError::LogNotConsistent {
+                    old_root: old.clone(),
+                    new_root: new.clone(),
+                })?;
+        if old == DynHash::from(old_hash) && new == DynHash::from(new_hash) {
+            Ok(())
+        } else {
+            Err(ProofError::LogNotConsistent {
+                old_root: old,
+                new_root: new,
+            })
+        }
     }
 
     /// Uploads package content to the registry.
--- a/crates/client/src/config.rs
+++ b/crates/client/src/config.rs
@@ -68,6 +68,9 @@ pub struct Config {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub default_url: Option<String>,
 
+    /// List of registries
+    pub registries: Vec<String>,
+
     /// The path to the directory where per-registry packages are stored.
     ///
     /// This path is expected to be relative to the configuration file.
@@ -75,7 +78,7 @@ pub struct Config {
     /// If `None`, the default of `$CACHE_DIR/warg/packages` is used, where
     /// `$CACHE_DIR` is the platform-specific cache directory.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub packages_dir: Option<PathBuf>,
+    pub registries_dir: Option<PathBuf>,
 
     /// The path to the directory where package content is stored.
     ///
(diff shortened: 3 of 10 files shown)


macovedj:
So far this PR just stores checkpoints anytime they're updated so that and old root and new root can be used to hit the consistency endpoint that the warg server offers.  Hoping to finalize and use the response to evaluate a proof soon.

crates/client/src/storage/fs.rs
```diff
@@ -118,6 +121,16 @@ impl PackageStorage for FileSystemPackageStorage {
         store(&self.package_path(&info.name), info).await
     }
 
+    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<MapCheckpoint>>> {
+        let checkpoint = load(&self.package_path("checkpoint")).await?;
```
Kylebrown9, line 125: I don't think this is the correct place to put the checkpoint. This is the path for a package named `checkpoint`. 

This should probably be a new file path with its own function & hard-coded name like `pending_publish_path`.

crates/client/src/storage/fs.rs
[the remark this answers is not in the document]
macovedj: looks like this is the only one right now

Kylebrown9:
The proof/API side of this looks good, but I'm not sure about the storage/CLI part.

@peterhuene can you take a look at it?

crates/client/src/config.rs
```diff
@@ -85,6 +85,15 @@ pub struct Config {
...
     pub content_dir: Option<PathBuf>,
+
+    /// The path to the directory where package content is stored.
+    ///
+    /// This path is expected to be relative to the configuration file.
+    ///
+    /// If `None`, the default of `$CACHE_DIR/warg/checkpoint` is used, where
+    /// `$CACHE_DIR` is the platform-specific cache directory.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub checkpoint_path: Option<PathBuf>,
```
peterhuene, line 96: I don't think this is correct. The checkpoint should be stored per-registry, and this appears to be a single checkpoint for the client regardless of registry.

Perhaps `packages_dir` should become `registries_dir` (`$CACHE_DIR/warg/registries`), which underneath it has:

* `packages` - a directory containing the package logs
* `checkpoint` - a file containing the last known registry checkpoint
* `operator.log` - the registry's operator log (for the future, don't need to implement this now as we're not doing anything with the operator log currently).

crates/client/src/config.rs
```diff
@@ -228,10 +239,24 @@ impl Config {
             })
     }
 
+    /// Gets the path to the file where the previously fetched checkpoint is stored
+    pub fn checkpoint_path(&self) -> Result<PathBuf> {
```
peterhuene, line 243: As a consequence of the above, I'd expect this method to be removed.

crates/client/src/lib.rs
```diff
@@ -33,20 +34,22 @@ pub mod storage;
...
-pub struct Client<P, C> {
+pub struct Client<P, C, T> {
     packages: P,
     content: C,
     api: api::Client,
+    checkpoint: T,
 }
 
-impl<P: PackageStorage, C: ContentStorage> Client<P, C> {
+impl<P: PackageStorage, C: ContentStorage, T: CheckpointStorage> Client<P, C, T> {
```
peterhuene, line 44: I think we should probably rename `PackageStorage` to `RegistryStorage` and incorporate the checkpoint into it; eventually we'll need to add in operator log reading/writing into it and I'd rather not have more traits.

The reason there's two traits here is that content storage is across all registries, whereas the rest of the data is per-registry.

peterhuene (commented):
Thanks so much for taking these changes on, @macovedj!

I have some feedback comments to address. Most of the feedback stems from the existing design needed to be refactored to accommodate more information about this and future needs.

crates/client/src/config.rs
macovedj: Ah this resonates.  I was actually looking into this [client operator issue](https://github.com/bytecodealliance/registry/issues/83) a bit this morning, and was thinking it would require something like what you're suggesting.  I'll try and have some changes up soon.

crates/client/src/api.rs
```diff
@@ -219,9 +225,48 @@ impl Client {
         old_root: DynHash,
         new_root: DynHash,
     ) -> ProofResult<()> {
-        dbg!(old_root);
-        dbg!(new_root);
-        todo!()
+        let old = old_root.clone();
+        let new = new_root.clone();
+        let request = ConsistencyRequest { old_root, new_root };
```
peterhuene, line 230: Side note: the request types in the API crate should borrow their data so we don't have these unnecessary clones.

I'll file an issue.

peterhuene: Filed as https://github.com/bytecodealliance/registry/issues/104

macovedj:
@peterhuene I think that I addressed most of the comments made.  For the time being, I didn't see an alternative to hardcoding the registries that we expect to interact with and want to be a part of local storage.  I'm thinking that the eventual solution is to have the server responses include something like a `registry_name` that would be used to determine which subfolder package information would get stored in, but wasn't sure whether or not that should get scoped into this PR.

crates/client/src/config.rs
```diff
@@ -60,6 +60,15 @@ fn normalize_path(path: &Path) -> PathBuf {
     ret
 }
 
+/// Paths used for storage
+pub struct StoragePaths {
+    /// url path
+    pub url: Url,
+    /// registries path
```
peterhuene, line 67:
```suggestion
    /// The path to the registry storage directory.
```

crates/client/src/config.rs
```diff
@@ -60,6 +60,15 @@ fn normalize_path(path: &Path) -> PathBuf {
...
+    pub registries_dir: PathBuf,
+    /// content path
```
peterhuene, line 69:
```suggestion
    /// The path to the content storage directory.
```

crates/client/src/config.rs
```diff
@@ -68,14 +77,17 @@ pub struct Config {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub default_url: Option<String>,
 
+    /// List of registries
+    pub registries: Vec<String>,
```
peterhuene, line 81: I can't seem to determine what this is used for. Could you describe the intention behind this field?

crates/client/src/config.rs
```diff
@@ -68,14 +77,17 @@ pub struct Config {
...
+
     /// The path to the directory where per-registry packages are stored.
```
peterhuene, line 83:
```suggestion
    /// The path to the top-level directory where per-registry information is stored.
```

crates/client/src/config.rs
```diff
@@ -68,14 +77,17 @@ pub struct Config {
...
     ///
     /// This path is expected to be relative to the configuration file.
     ///
     /// If `None`, the default of `$CACHE_DIR/warg/packages` is used, where
```
peterhuene, line 87:
```suggestion
    /// If `None`, the default of `$CACHE_DIR/warg/registries` is used, where
```

crates/client/src/config.rs
```diff
@@ -201,8 +214,8 @@ impl Config {
     }
 
     /// Gets the path to the directory where per-registry packages are stored.
-    pub fn packages_dir(&self) -> Result<PathBuf> {
-        self.packages_dir
+    pub fn registries_dir(&self) -> Result<PathBuf> {
```
peterhuene, line 217: Can we update the default in here from `warg/packages` to `warg/registries`?

crates/client/src/config.rs
```diff
@@ -231,15 +244,19 @@ impl Config {
...
-    ) -> Result<(Url, PathBuf, PathBuf), ClientError> {
+    ) -> Result<StoragePaths, ClientError> {
         let url = api::Client::validate_url(
             url.or(self.default_url.as_deref())
                 .ok_or(ClientError::NoDefaultUrl)?,
         )?;
 
         let host = url.host().unwrap().to_string().to_ascii_lowercase();
-        let packages_dir = self.packages_dir()?.join(host);
+        let registries_dir = self.registries_dir()?.join(host);
```
peterhuene, line 254: As a note, this is the line where we hand out the per-registry storage location based on the host of the registry we're communicating with.

crates/client/src/lib.rs
```diff
@@ -71,12 +72,12 @@ impl<P: PackageStorage, C: ContentStorage> Client<P, C> {
     pub async fn publish(&self, signing_key: &signing::PrivateKey) -> ClientResult<()> {
         let info = self
             .packages
-            .load_publish()
+            .load_publish("dogfood")
```
peterhuene, line 75: The `RegistryStorage` is already per-registry, so no need for a hardcoded `"dogfood"` here and below (see my comments on the `RegistryStorage` trait).

crates/client/src/lib.rs
```diff
@@ -39,7 +40,7 @@ pub struct Client<P, C> {
     api: api::Client,
 }
 
-impl<P: PackageStorage, C: ContentStorage> Client<P, C> {
+impl<P: RegistryStorage, C: ContentStorage> Client<P, C> {
```
peterhuene, line 43: There's a few places here where comments refer to "package" storage that should be updated to "registry storage".

crates/client/src/storage.rs
```diff
@@ -24,27 +24,38 @@ pub use fs::*;
 /// Package storage data must be synchronized if shared between
 /// multiple threads and processes.
 #[async_trait]
-pub trait PackageStorage: Send + Sync {
+pub trait RegistryStorage: Send + Sync {
+    ///  Loads most recent checkpoint
+    async fn load_checkpoint(&self, registry: &str)
```
peterhuene, line 29: The `RegistryStorage` is already specific to a particular registry; we should revert any changes to the existing methods from the previous `PackageStorage` trait.

We should also update the comments on the trait to something like:

```
/// Trait for registry storage implementations.
///
/// Stores information such as package/operator logs and checkpoints
/// on a per-registry basis.
///
/// Registry storage data must be synchronized if shared between
/// multiple threads and processes.
```

macovedj:
On second thought, I think I can get around the hardcoding for the time being, but still have open questions around whether registry identifiers need to be included in endpoint responses.  Should be able to push something to remove the hardcoding tomorrow.

peterhuene (commented):
The updated changes look like a step in the right direction.

I think we can remove the hardcoded "dogfood" references (see individual review comments).

crates/client/src/lib.rs
peterhuene: Also, should rename the `packages` method (perhaps simply to `registry`?).

macovedj:
Ok I think it's ready for review again.
One final thing with respect to this comment:

> I don't think this is correct. The checkpoint should be stored per-registry, and this appears to be a single checkpoint for the client regardless of registry.
> 
> Perhaps `packages_dir` should become `registries_dir` (`$CACHE_DIR/warg/registries`), which underneath it has a sub-directory per registry (e.g. `registry.bytecodealliance.org`) containing:
> 
> * `packages` - a directory containing the package logs for the registry
> * `checkpoint` - a file containing the last known registry checkpoint
> * `operator.log` - the registry's operator log (for the future, don't need to implement this now as we're not doing anything with the operator log currently).

The current changes in this PR rename `packages` to `registries`, and the only difference in the structure of its contents is that in a host subfolder, there's now a file called `checkpoint` which will store the content from the most recent checkpoint update for that host/registry server.  This didn't require any changes to any structs, just added a `.join("checkpoint")` in the new methods.  Since the methods for `operator.log` don't exist yet, and since a new struct wasn't necessary, there wasn't a need for any operator-related changes to appear in this PR, but I'm guessing that we'll see changes in the future PR that handles operator verification on the client.

I considered adding some reference to a a checkpoint path to to the `StoragePaths` struct, but it seemed like it would complicate things more than it would help.

crates/client/src/config.rs
Kylebrown9: Bump @macovedj

macovedj: Peter's comment helped me realize it wasn't necessary.  Went ahead and deleted this field.

crates/api/src/proof.rs
```diff
@@ -91,6 +91,15 @@ pub enum ProofError {
         /// The id of the package.
         id: LogId,
     },
+
+    /// Failed to prove consistency of a package.
```
peterhuene, line 95:
```suggestion
    /// Failed to prove consistency of the registry between two checkpoints.
```

crates/client/src/config.rs
```diff
@@ -60,6 +60,15 @@ fn normalize_path(path: &Path) -> PathBuf {
...
 
+/// Paths used for storage
+pub struct StoragePaths {
+    /// The registry URL relating to the storage paths.
+    pub url: Url,
+    /// The path to the registry storage directory.
+    pub registries_dir: PathBuf,
+    /// The path to the content storage directory.
+    pub content_dir: PathBuf,
+}
```
peterhuene, line 71: Nit: 

```suggestion
}

```

crates/client/src/storage.rs
```diff
@@ -19,12 +19,21 @@ use warg_protocol::{
...
+///
+/// Registry storage data must be synchronized if shared between
 /// multiple threads and processes.
 #[async_trait]
-pub trait PackageStorage: Send + Sync {
+pub trait RegistryStorage: Send + Sync {
+    ///  Loads most recent checkpoint
+    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<MapCheckpoint>>>;
+
+    ///  Stores most recent checkpoint
```
peterhuene, line 34: Nit:

```suggestion
    /// Stores most recent checkpoint
```

crates/client/src/storage.rs
```diff
@@ -19,12 +19,21 @@ use warg_protocol::{
...
-/// Package storage data must be synchronized if shared between
+/// Stores information such as package/operator logs and checkpoints
+/// on a per-registry basis.
+///
+/// Registry storage data must be synchronized if shared between
 /// multiple threads and processes.
 #[async_trait]
-pub trait PackageStorage: Send + Sync {
+pub trait RegistryStorage: Send + Sync {
+    ///  Loads most recent checkpoint
```
peterhuene, line 31: Nit:

```suggestion
    /// Loads most recent checkpoint
```

crates/client/src/storage/fs.rs
```diff
@@ -76,7 +79,16 @@ impl FileSystemPackageStorage {
...
-impl PackageStorage for FileSystemPackageStorage {
+impl RegistryStorage for FileSystemPackageStorage {
+    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<MapCheckpoint>>> {
+        let contents = load(&self.base_dir.join("checkpoint")).await;
+        contents
+    }
+
+    async fn store_checkpoint(&self, checkpoint: SerdeEnvelope<MapCheckpoint>) -> Result<()> {
+        store(&self.base_dir.join("checkpoint"), checkpoint).await;
+        Ok(())
```
peterhuene, line 90: This appears to be an unused result:

```suggestion
        store(&self.base_dir.join("checkpoint"), checkpoint).await
```

crates/client/src/storage/fs.rs
```diff
@@ -76,7 +79,16 @@ impl FileSystemPackageStorage {
 }
 
 #[async_trait]
-impl PackageStorage for FileSystemPackageStorage {
+impl RegistryStorage for FileSystemPackageStorage {
+    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<MapCheckpoint>>> {
+        let contents = load(&self.base_dir.join("checkpoint")).await;
+        contents
```
peterhuene, line 85: Nit (as it's not really the contents, just the result):

```suggestion
        load(&self.base_dir.join("checkpoint")).await
```

src/commands/config.rs
```diff
@@ -12,7 +12,7 @@ pub struct ConfigCommand {
 
     /// The path to the packages directory to use.
     #[clap(long, value_name = "PACKAGES")]
-    pub packages_dir: Option<PathBuf>,
+    pub registries_dir: Option<PathBuf>,
```
peterhuene, line 15: Need to update the comment on this field and the value to `REGISTRIES`.

peterhuene (commented):
Looks great!

Thanks for putting up with all my comments.

The remaining is almost entirely nits at this point, except for one unused result warning that needs fixing.

macovedj:
> Looks great!
> 
> Thanks for putting up with all my comments.
> 
> The remaining is almost entirely nits at this point, except for one unused result warning that needs fixing.

Appreciate it!  Thanks for putting up with my missing some of the comments that needed changing and such.  I'll be more vigilant on the next PR to avoid us having so much to cover.

crates/client/src/storage.rs
```diff
@@ -19,12 +19,21 @@ use warg_protocol::{
...
+/// Registry storage data must be synchronized if shared between
 /// multiple threads and processes.
 #[async_trait]
-pub trait PackageStorage: Send + Sync {
+pub trait RegistryStorage: Send + Sync {
+    /// Loads most recent checkpoint
+    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<MapCheckpoint>>>;
+
+    /// Stores most recent checkpoint
+    async fn store_checkpoint(&self, checkpoint: SerdeEnvelope<MapCheckpoint>);
```
peterhuene, line 35: I'd expect `store_checkpoint` to still return `Result<()>` as it is a fallible operation.

macovedj: Brought back the return type

src/commands/config.rs
peterhuene: I think the `value_name` needs updating still and that's the last of it 👍

peterhuene (approved):
Looks great 👍

Thanks for implementing this!

PR #103, opened 2023-05-04
Regex protoc version check

### LINKED ISSUE #86: Check in `build.rs` if the required `protoc` version is available
Use [`protoc_from_env`](https://docs.rs/prost-build/latest/prost_build/fn.protoc_from_env.html) to find the `protoc` that prost will use and then run the version command to check if version `3.15` or greater is available.

Gearme:
Fixes #86 using `Command` for `protoc --version` and the `regex` crate on a modified template from semver.org to check for a minimum major and minor version of protoc in both the old and the new versioning schema of the protobuf project (see [protobuf's versioning scheme documentation](https://protobuf.dev/support/version-support/)).

Causes the `transparency` and `protocol` crates' build scripts to panic with a helpful error message if the specified minimum `protoc` version is not satisfied.


crates/protocol/build.rs
```diff
@@ -28,3 +30,32 @@ fn main() -> Result<()> {
...
+fn verify_protoc_version(min_major: u32, min_minor: u32) {
+    let protoc_path = prost_build::protoc_from_env();
+    let protoc_version_output = Command::new(protoc_path)
+        .args(["--version"])
+        .output()
+        .unwrap();
+    let protoc_version = String::from_utf8(protoc_version_output.stdout).unwrap();
+
+    // based on semver.org's recommended regex, modified for lib name and optional patch capture group to accomodate libprotoc's versioning scheme
+    let re = Regex::new(r"^libprotoc (?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.?(?P<patch>0|[1-9]\d*)?(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$").unwrap();
```
Kylebrown9, line 45: Any reason to hard-code a regex instead of using [`semver::Version::parse`](https://docs.rs/semver/1.0.17/semver/struct.Version.html#method.parse)?

Gearme: Yes, namely encountering the very first error case mentioned in the link you gave ;) Though I admit I actually assumed it simply didn't support fewer than three segments full stop - the docs seem to indicate that one could use VersionReq instead. Since semver is already in the deps anyway, I'll give that a try and see if it's a little more elegant.

Kylebrown9: Interesting, it looks like according to [semver.org](https://semver.org/) the `semver` crate is correct.

```bnf
<version core> ::= <major> "." <minor> "." <patch>
```

Kylebrown9: That's sort of a moot point though, we do need to parse these whether they count as semantic or not and if we need a regex that's fine.

Kylebrown9:
I'll test this on a couple systems in the next few days and approve after.

Kylebrown9 (approved):

PR #107, opened 2023-05-08
Client Operator Validation

macovedj:
Implementation of [Operator Validations](https://github.com/bytecodealliance/registry/issues/83)

Currently just writing some information from operator logs to file system storage.  Validation logic is coming

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -343,7 +343,7 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             if let Err(error) = operator_info.state.validate(&record) {
                 error!("failed to validate: {}", error);
             }
-            self.registry.store_operator(record).await?;
+            self.registry.store_operator(operator_info).await?;
         }
         let mut heads = Vec::with_capacity(packages.len());
         for (name, records) in response.packages {
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -382,7 +382,7 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             operator
                 .state
                 .validate(&record)
-                .map_err(|error| ClientError::OperatorValidationFailed { error })?;
+                .map_err(|inner| ClientError::OperatorValidationFailed { inner })?;
             if let Some(head) = operator.state.head() {
                 heads.push(LogLeaf {
                     log_id: LogId::operator_log::<Sha256>(),
@@ -559,7 +559,7 @@ pub enum ClientError {
     #[error("operator failed validation")]
     OperatorValidationFailed {
         /// The validation error.
-        error: operator::ValidationError,
+        inner: operator::ValidationError,
     },
     /// The package already exists and cannot be initialized.
     #[error("package `{package}` already exists and cannot be initialized")]
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -556,7 +556,7 @@ pub enum ClientError {
     NoDefaultUrl,
 
     /// The operator failed validation.
-    #[error("operator failed validation")]
+    #[error("operator failed validation: {inner}")]
     OperatorValidationFailed {
         /// The validation error.
         inner: operator::ValidationError,
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -343,21 +343,6 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             .await?;
 
         let mut heads = Vec::with_capacity(packages.len());
-        let operator_records = response.operator;
-        for record in operator_records {
-            let record: ProtoEnvelope<operator::OperatorRecord> = record.try_into()?;
-            let mut operator_info = storage::OperatorInfo::default();
-            if let Err(error) = operator_info.state.validate(&record) {
-                error!("failed to validate: {}", error);
-            }
-            if let Some(head) = operator_info.state.head() {
-                heads.push(LogLeaf {
-                    log_id: LogId::operator_log::<Sha256>(),
-                    record_id: head.digest.clone(),
-                })
-            }
-            self.registry.store_operator(operator_info).await?;
-        }
         for (name, records) in response.packages {
             match packages.get_mut(&name) {
                 Some(package) => {
@@ -386,13 +371,6 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             }
         }
 
-        self.api.prove_inclusion(checkpoint.as_ref(), heads).await?;
-
-        for package in packages.values_mut() {
-            package.checkpoint = Some(checkpoint.clone());
-            self.registry.store_package(package).await?;
-        }
-
         let old_checkpoint = self.registry.load_checkpoint().await?;
         if let Some(cp) = old_checkpoint {
             let old_cp = cp.as_ref().clone();
@@ -401,6 +379,29 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
                 .prove_log_consistency(old_cp.log_root, new_cp.log_root)
                 .await?;
         }
+
+        let operator_records = response.operator;
+        for record in operator_records {
+            let record: ProtoEnvelope<operator::OperatorRecord> = record.try_into()?;
+            let mut operator_info = storage::OperatorInfo::default();
+            if let Err(error) = operator_info.state.validate(&record) {
+                error!("failed to validate: {}", error);
+            }
+            if let Some(head) = operator_info.state.head() {
+                heads.push(LogLeaf {
+                    log_id: LogId::operator_log::<Sha256>(),
+                    record_id: head.digest.clone(),
+                })
+            }
+            self.registry.store_operator(operator_info).await?;
+        }
+        self.api.prove_inclusion(checkpoint.as_ref(), heads).await?;
+
+        for package in packages.values_mut() {
+            package.checkpoint = Some(checkpoint.clone());
+            self.registry.store_package(package).await?;
+        }
+
         self.registry.store_checkpoint(checkpoint.clone()).await?;
 
         Ok(())
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -24,7 +24,7 @@ use warg_crypto::{
 };
 use warg_protocol::{
     operator,
-    package::{self, ValidationError},
+    package::{self},
     registry::{LogId, LogLeaf, MapCheckpoint},
     ProtoEnvelope, SerdeEnvelope, Version, VersionReq,
 };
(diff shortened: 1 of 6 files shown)


crates/client/src/lib.rs
```diff
@@ -335,6 +337,21 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             .await?;
 
         let mut heads = Vec::with_capacity(packages.len());
+        let operator_records = response.operator;
```
peterhuene, line 340: So this looks relatively correct for validating operator records, there's a few things that need addressing here:

* The `FetchRequest` above is still passing a last operator record of `None`, which means the response will have _every_ record in the operator log that's in the given checkpoint, starting from the beginning of the log.

* This validation only works since we're assuming every record is being returned, which is why we start from a default `OperatorInfo`.

The way I would expect this to work is that, before fetching the logs, we load the operator info from storage (or create a default one). We pass the last known operator record identifier in `FetchRequest` (or `None` if the head is `None`). We then validate the records returned in the response.

Only after all validation passes should we persist the latest log states (for both packages and the operator log). That said, we should probably move the call to `prove_log_consistency` to above where we store the package logs and the code to store the operator log to below where we store the package logs.

crates/client/src/storage.rs
```diff
@@ -34,6 +35,14 @@ pub trait RegistryStorage: Send + Sync {
     /// Stores most recent checkpoint
     async fn store_checkpoint(&self, checkpoint: SerdeEnvelope<MapCheckpoint>) -> Result<()>;
 
+    /// Loads the operator information from the storage.
+    ///
+    /// Returns `Ok(None)` if the information is not present.
+    async fn load_operator(&self) -> Result<Option<OperatorRecord>>;
```
peterhuene, line 41: This doesn't seem to be called. Additionally, I'd expect this to return `OperatorInfo`.

crates/protocol/src/operator/model.rs
```diff
@@ -7,7 +7,7 @@ use warg_crypto::hash::{DynHash, HashAlgorithm};
 use warg_crypto::signing;
 
 /// A operator record is a collection of entries published together by the same author
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
peterhuene, line 10: I don't think changes to this file are necessary because we shouldn't be serializing records on the client, only the validation state.

crates/protocol/src/proto_envelope.rs
```diff
@@ -11,7 +11,7 @@ use warg_crypto::{hash::DynHashError, signing, Decode, Signable};
 /// The envelope struct is used to keep around the original
 /// bytes that the content was serialized into in case
 /// the serialization is not canonical.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
peterhuene, line 14: Likewise, the change should not be necessary (and this type is intentionally not serializable).

crates/protocol/src/operator/mod.rs
```diff
@@ -4,7 +4,7 @@ use prost::Message;
 use thiserror::Error;
 use warg_crypto::{hash::DynHash, Decode, Encode, Signable};
 
-mod model;
+pub mod model;
```
peterhuene, line 7: Given the other changes to this module aren't necessary, I think we can back this change out.

crates/client/src/storage/fs.rs
```diff
@@ -120,6 +121,45 @@ impl RegistryStorage for FileSystemPackageStorage {
         Ok(packages)
     }
 
+    async fn load_operator(&self) -> Result<Option<OperatorRecord>> {
```
peterhuene, line 124: As stated earlier, if this were to be loading an `OperatorInfo`, I don't think the changes to make `OperatorRecord` serializable would be necessary.

crates/client/src/storage/fs.rs
```diff
@@ -120,6 +121,45 @@ impl RegistryStorage for FileSystemPackageStorage {
...
+        serde_json::from_str(&contents).with_context(|| {
+            format!(
+                "failed to deserialize contents of `{path}`",
+                path = path.display()
+            )
+        })
+    }
+
+    async fn store_operator(&self, operator: OperatorInfo) -> Result<()> {
+        let path = &self.base_dir.join("operator.log");
```
peterhuene, line 143: Could the body of this function look more like `store_package`?

That is:

```rust
store(&self.operator_path(), info).await
```

crates/client/src/storage/fs.rs
peterhuene: I'd also expect the implementation to look more like:

```rust
Ok(load(&self.operator_path()).await?)
```

peterhuene (commented):
@macovedj This looks like it's on track; just some comments to address.

macovedj:
@peterhuene I'll try and address these soon.  I also was looking at a pattern that is pretty common in [this file](https://github.com/bytecodealliance/registry/blob/main/crates/client/src/lib.rs).  I was trying to decided where the best logical place is to load the operator info, and noticed that we almost always load packages and then pass them as an argument to `update_checkpoint`.  Was wondering if you support what I think would be a simplification, making the code more DRY, where `update_checkpoint` instead takes a list of package names as an arg, rather than a list of package info, and does the package loading inside of the function call.  This should reduce the number of places where `load_package` has to get called in the code.

peterhuene:
I'm fine with refactoring `update_checkpoint` in the way you describe as long as we ensure we don't have to load a package information more than once per operation (thus it would have to return the package information as some of the operations make use of it after the checkpoint update).

crates/client/src/lib.rs
```diff
@@ -378,6 +379,29 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
                 .prove_log_consistency(old_cp.log_root, new_cp.log_root)
                 .await?;
         }
+
+        let operator_records = response.operator;
+        for record in operator_records {
+            let record: ProtoEnvelope<operator::OperatorRecord> = record.try_into()?;
+            let mut operator_info = storage::OperatorInfo::default();
+            if let Err(error) = operator_info.state.validate(&record) {
+                error!("failed to validate: {}", error);
```
peterhuene, line 388: This should probably return a `ClientError::OperatorValidationFailed` error instead of just logging an error.

crates/client/src/lib.rs
```diff
@@ -378,6 +379,29 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+            }
+            if let Some(head) = operator_info.state.head() {
+                heads.push(LogLeaf {
+                    log_id: LogId::operator_log::<Sha256>(),
+                    record_id: head.digest.clone(),
+                })
+            }
+            self.registry.store_operator(operator_info).await?;
```
peterhuene, line 396: Storing the operator log state should be moved out of the loop and after proving inclusion.

crates/client/src/lib.rs
```diff
@@ -378,6 +379,29 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
                 .prove_log_consistency(old_cp.log_root, new_cp.log_root)
                 .await?;
         }
+
+        let operator_records = response.operator;
+        for record in operator_records {
+            let record: ProtoEnvelope<operator::OperatorRecord> = record.try_into()?;
+            let mut operator_info = storage::OperatorInfo::default();
```
peterhuene, line 386: This should be removed and instead use the operator information loaded previously (there should be one load and one save).

crates/client/src/lib.rs
```diff
@@ -304,6 +306,12 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
             map_root = checkpoint.as_ref().map_root
         );
 
+        let operator = self
+            .registry
+            .load_operator()
+            .await?
+            .map(|op| op.state.head().as_ref().map(|h| h.digest.clone()).unwrap());
```
peterhuene, line 313: I think this should be:

```rust
let operator = self
    .registry
    .load_operator()
    .await?
    .unwrap_or_default();
```

The reason being is that we want to use the same operator validation state we might have stored previously when validating new records for the operator log.

And then where we use it in `FetchRequest`:

```rust
operator: operator.state.head().as_ref().map(|h| h.digest.clone())
```

Note that it's valid for this to be `None` if we created a default operator info above (i.e. we haven't stored any state, so start from the beginning).

Finally, where we do the operator log validation, we should updating `operator` and not creating a new operator information every time; there should also be a single save of the operator info.

crates/client/src/storage.rs
```diff
@@ -90,6 +99,28 @@ pub trait ContentStorage: Send + Sync {
     ) -> Result<DynHash>;
 }
 
+/// Represents information about a registry operator.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorInfo {
+    /// The last known checkpoint of the package.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub checkpoint: Option<SerdeEnvelope<MapCheckpoint>>,
```
peterhuene, line 108: I don't think we need to store the checkpoint here. We store it per-package because not every package log is updated at once, whereas the operator log will always be updated with every request to update any package.

```suggestion
    /// The last known checkpoint of the package.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub checkpoint: Option<SerdeEnvelope<MapCheckpoint>>,
```

crates/client/src/storage.rs
```diff
@@ -90,6 +99,28 @@ pub trait ContentStorage: Send + Sync {
...
+    /// The current validation state of the package.
```
peterhuene, line 109:
```suggestion
    /// The current validation state of the operator.
```

crates/client/src/storage.rs
```diff
@@ -90,6 +99,28 @@ pub trait ContentStorage: Send + Sync {
...
+
+impl OperatorInfo {
+    /// Creates a new operator info for the given operator name and url.
+    pub fn default() -> Self {
+        Self {
+            checkpoint: None,
+            state: operator::Validator::default(),
+        }
+    }
+}
```
peterhuene, line 122: Could we instead put a derive of `Default` on `OperatorInfo`?

```suggestion
```

crates/client/src/lib.rs
```diff
@@ -559,6 +554,9 @@ pub enum ClientError {
     #[error("no default registry server URL is configured")]
     NoDefaultUrl,
 
+    /// The operator failed validation.
+    #[error("operator failed validation")]
+    OperatorValidationFailed,
```
peterhuene, line 559: This should probably store the validation error so it can be printed with the message, e.g. `operator validation failed: {0}`.

peterhuene (commented):
Just the one last comment, looks great!

crates/client/src/lib.rs
```diff
@@ -535,6 +555,12 @@ pub enum ClientError {
     #[error("no default registry server URL is configured")]
     NoDefaultUrl,
 
+    /// The operator failed validation.
+    #[error("operator failed validation")]
```
peterhuene, line 559: Oh whoops, approved too soon; can you update the error message to `operator validation failed: {inner}`?

peterhuene (approved):

peterhuene:
👍 Thanks for implementing this @macovedj.

PR #108, opened 2023-05-09
feat: add signing key operations

fibonacci1729:
Using @peterhuene's work in `cargo-component`, this PR adds support for creating, updating, and deleting signing keys using `keyring`.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -8,6 +8,7 @@ use warg_client::{ClientError, Config, FileSystemClient, StorageLockResult};
 mod config;
 mod download;
 mod info;
+mod key;
 mod publish;
 mod run;
 mod update;
@@ -15,6 +16,7 @@ mod update;
 pub use self::config::*;
 pub use self::download::*;
 pub use self::info::*;
+pub use self::key::*;
 pub use self::publish::*;
 pub use self::run::*;
 pub use self::update::*;
@@ -25,6 +27,9 @@ pub struct CommonOptions {
     /// The URL of the registry to use.
     #[clap(long, value_name = "URL")]
     pub registry: Option<String>,
+    /// The name to use for the signing key.
+    #[clap(long, short, value_name = "KEY_NAME", default_value = "default")]
+    pub key_name: String,
     /// The path to the client configuration file to use.
     ///
     /// If not specified, the following locations are searched in order: `./warg-config.json`, `<system-config-dir>/warg/config.json`.
--- a/src/commands/publish.rs
+++ b/src/commands/publish.rs
@@ -1,24 +1,17 @@
 use super::CommonOptions;
+use crate::signing::get_signing_key;
 use anyhow::{anyhow, bail, Context, Result};
 use clap::{Args, Subcommand};
 use futures::TryStreamExt;
-use std::{env, future::Future, path::PathBuf};
+use std::{future::Future, path::PathBuf};
 use tokio::io::BufReader;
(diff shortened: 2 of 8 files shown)


src/commands.rs
```diff
@@ -35,6 +43,42 @@ pub struct CommonOptions {
 }
 
 impl CommonOptions {
+    /// Get the signing key for the specified user and registry.
+    pub fn get_signing_key(&self) -> Result<signing::PrivateKey> {
```
fibonacci1729, line 47: This probably needs a closer look; i'm not sure i am using the `default_url` from the `Config` correctly here.

peterhuene: I think we can probably remove this method (see my other comment).

src/commands/publish.rs
```diff
@@ -58,9 +50,12 @@ where
 }
 
 /// Submits a publish to the registry.
-async fn submit(client: &FileSystemClient, info: PublishInfo) -> Result<()> {
-    let signing_key = demo_signing_key()?;
-    client.publish_with_info(&signing_key, info).await?;
+async fn submit(
+    client: &FileSystemClient,
+    info: PublishInfo,
+    signing_key: &PrivateKey,
```
peterhuene, line 56: Instead of taking the signing key here, if we instead pass in the username, we could call

```rust
get_signing_key(user, host)
```

where `host` could be derived from `client.url()`?

That would eliminate the need for `CommonOptions::get_signing_key`.

It would also delay any interactions with the key store until right before we know we're submitting a publish operation.

peterhuene (commented):
Looks great overall! Just a few comments.

src/commands/publish.rs
fibonacci1729: @Kylebrown9 the user name associated with the signing-key (correlated via `warg key add ...`)

peterhuene: `cargo-component` uses the same strategy so that one signing key isn't used per-registry, but the CLI user gets to decide which key to use for that registry to publish with.

Perhaps a better name is needed? Like "key name" or some such? I agree that "user name" could be confused with some sort of auth for the registry itself.

peterhuene (approved):
Other than the discussion around "user name" that needs to be resolved, the changes LGTM 👍

src/commands/publish.rs
fibonacci1729: `--key-name` (or `--key`?) sounds good to me; @Kylebrown9 do you have a preference?

fibonacci1729: I updated this PR to use `--key-name` instead of `--user`.

src/commands/publish.rs
```diff
@@ -304,6 +312,12 @@ pub struct PublishAbortCommand {
     /// The common command options.
     #[clap(flatten)]
     pub common: CommonOptions,
+    /// The user name to use for the signing key.
+    #[clap(long, short, value_name = "USER", default_value = "default")]
+    pub user: String,
+    /// The host name of the registry to set the signing key for.
+    #[clap(value_name = "HOST")]
+    pub host: String,
```
peterhuene, line 320: Is this needed?

peterhuene (approved):
Other than the question regarding the abort command's options, I think this is good to go.

src/commands/publish.rs
fibonacci1729: Ah good catch!

peterhuene (approved):

PR #109, opened 2023-05-11
fix: correct data_store server config debug fmt

### LINKED ISSUE #110: Incorrect data_store server config debug fmt
Should be `shutdown` -> `data_store`:

https://github.com/bytecodealliance/registry/blob/3d9931109c7ce37dc2f80c51c457a128282060c5/crates/server/src/lib.rs#L46-L49

kriswuollett:
- fixes #110


peterhuene (approved):
Nice catch!

PR #113, opened 2023-05-12
fix: return bound endpoints after server started

### LINKED ISSUE #112: Cleanup: remove server partial move of self in `run`
It may make more sense to split `run` into `start` and `join` and merge `bind` into `start` in order to have a clean way to give locally bound ports to tests from a `start` method which also helps guarantee in a way that the web server has started serving on the bound port with a listener before tests hit it.

kriswuollett:
The impetus behind the change is to support adding monitoring endpoints with an optional additional TCP listener. While refactoring the server it was discovered that the run method moves the Server self, and so fixed #112 to also ensure that the API endpoint was running before tests could discover the listener's locally bound port.

In addition a common cancellation token is added to the server to propagate shutdown to both the core service as well as the API server endpoint. It can be later used to support additional services later such as a monitoring endpoint(s).

### DIFF
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -12,6 +12,8 @@ use std::{
     sync::Arc,
     time::Duration,
 };
+use tokio::task::JoinSet;
+use tokio_util::sync::CancellationToken;
 use tower_http::{
     trace::{DefaultMakeSpan, DefaultOnResponse, TraceLayer},
     LatencyUnit,
@@ -30,7 +32,7 @@ const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(5);
 
 /// The server configuration.
 pub struct Config {
-    operator_key: PrivateKey,
+    operator_key: Option<PrivateKey>,
     addr: Option<SocketAddr>,
     data_store: Option<Box<dyn datastore::DataStore>>,
     content_dir: Option<PathBuf>,
@@ -58,7 +60,7 @@ impl Config {
     /// Creates a new server configuration.
     pub fn new(operator_key: PrivateKey) -> Self {
         Self {
-            operator_key,
+            operator_key: Some(operator_key),
             addr: None,
             data_store: None,
             content_dir: None,
@@ -120,7 +122,15 @@ impl Config {
 /// Represents the warg registry server.
 pub struct Server {
     config: Config,
-    listener: Option<TcpListener>,
+    endpoints: Option<Endpoints>,
+    token: CancellationToken,
+    tasks: JoinSet<Result<()>>,
+}
+
+/// The bound endpoints for the warg registry server.
+#[derive(Clone)]
+pub struct Endpoints {
+    pub api: SocketAddr,
 }
 
 impl Server {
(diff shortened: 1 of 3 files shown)


kriswuollett:
Manually tested that `Ctrl-C` still works:

CleanShot 2023-05-12 at 11 42 52@2x

crates/server/src/lib.rs
```diff
@@ -120,60 +122,67 @@ impl Config {
...
-        let listener = self.listener.unwrap();
+            .context("failed to get local address for api endpoint listen socket")?;
+        tracing::debug!("api endpoint bound to address `{addr}`");
 
-        tracing::debug!(
-            "using server configuration: {config:?}",
-            config = self.config
-        );
+        let endpoints = Endpoints { api: addr };
+        self.endpoints = Some(endpoints.to_owned());
```
peterhuene, line 177: Nit (for clarity):

```suggestion
        self.endpoints = Some(endpoints.clone());
```

crates/server/src/lib.rs
```diff
@@ -183,32 +192,69 @@ impl Server {
...
-        if let Some(shutdown) = self.config.shutdown {
-            tracing::debug!("server is running with a shutdown signal");
+        // Shut down core service when token cancelled.
+        let token = self.token.clone();
+        self.tasks.spawn(async move {
+            token.cancelled().await;
+            tracing::info!("waiting for core service to stop");
+            handle.stop().await;
+            Ok(())
+        });
```
peterhuene, line 211: Could this task be merged with the next task? That is to say, after the `server` future completes in the next task, we simply wait on the stop handle for the core service.

Given this design, I would expect a 1:1 between a task in the task set and a running server (API in this case, but soon a monitoring server, etc). Each would shutdown in their own way.

crates/server/src/lib.rs
```diff
@@ -120,60 +122,67 @@ impl Config {
...
-    listener: Option<TcpListener>,
+    endpoints: Option<Endpoints>,
+    token: CancellationToken,
+    tasks: JoinSet<Result<()>>,
+}
+
+/// The bound endpoints for the warg registry server.
+#[derive(Clone)]
+pub struct Endpoints {
+    pub api: SocketAddr,
```
peterhuene, line 133:
```suggestion
    /// The address of the API endpoint.
    pub api: SocketAddr,
```

crates/server/src/lib.rs
```diff
@@ -120,60 +122,67 @@ impl Config {
...
             config,
-            listener: None,
+            token: CancellationToken::new(),
+            endpoints: None,
+            tasks: JoinSet::new(),
         }
     }
 
-    /// Binds the server to the configured address.
+    /// Starts the server and binds its endpoint to the configured address.
```
peterhuene, line 147:
```suggestion
    /// Starts the server and binds its endpoints to the configured addresses.
```

crates/server/src/lib.rs
```diff
@@ -120,60 +122,67 @@ impl Config {
...
     ///
-    /// Returns the address the server bound to.
-    pub fn bind(&mut self) -> Result<SocketAddr> {
+    /// Returns the endpoints the server bound to.
+    pub async fn start(&mut self) -> Result<Endpoints> {
+        assert!(
+            self.endpoints.is_none(),
+            "cannot start server multiple times"
+        );
```
peterhuene, line 154: So I'm personally not a big fan of taking `&mut self` for this as it leads to confusing internal state and needing to wrap things in `Option` just for this method to take their contents.

What if we renamed this `run` method to `spawn` and in addition to returning the endpoints, we return a type wrapping both the `JoinSet` representing the spawned server's tasks and the `CancellationToken` that can be used to stop the server. Thus, the type could have a `stop` method on it to stop the server and also implement `Future` that a caller can await on to join the tasks in the set?

peterhuene (commented):
@kriswuollett Thanks for refactoring this!

Overall, I think this makes a lot of sense and will make things cleaner once we have additional endpoints we want to run as part of the same "warg server".

I just have some comments to go over with you.

crates/server/src/lib.rs
Kylebrown9: Agree, I was just about to ask why required fields were becoming optional. I'd much rather have spawning produce a dedicated type that represents the info for a started service.

crates/server/src/lib.rs
kriswuollett: done

crates/server/src/lib.rs
kriswuollett: I'd think due to the nature of fanning out a cancelation signal chain, a cancellation token should be cloned for each owned depedency, somewhat similar to how [Go Context](https://pkg.go.dev/context) works. Otherwise there virtually will be for-loop boilerplate being written to cancel "children" all separately as a server grows. That allows things to shut down in parallel easily if needed.

But yes, could keep things simple for now and merge the shutdown tasks instead of renaming chaining of them through the tokens themselves. However I'd either want to use a broadcast channel instead, or an additional token for draining in #111.

kriswuollett (commented):
Reviewed comments, thanks! Primary pushback is the method of controlling shutdown.

crates/server/src/lib.rs
peterhuene: I don't think we want the core service to be stopped in parallel to the hyper server shutting down as it might race.

If we call `stop` on the core service, it will break out of its loop and the task will complete; note that the `CoreService` instance itself won't be dropped as a result of this as there are references to it from the hyper server.

If this is done in parallel to the hyper server shutting down, then it's possible a request being drained attempts to enqueue work for the core service to do; if it so happens that the core service already stopped and the send limit for the message queue is reached, it'll block waiting for the core service to remove a message which will never occur since its task has ended.

We also don't want to close the message queue's sender as part of stopping the core service as this would result in a panic (even a graceful error back to the client here is not ideal) and a reset of the connection for any request attempting to enqueue work for the core service to do after it has stopped.

I believe the correct order of operations here is to shutdown the hyper server then join on the core service, like we had before.

peterhuene: Note that eventually the core service won't be part `warg-server`, but instead another backend service for it to talk to to process new record entries; it's the limiting factor of why we can't spin up multiple `warg-server` instances right now.

crates/server/src/lib.rs
peterhuene: The existing code for `bind` method is `&mut self`, but the `run` method was `self`, both are now combined into your `start` method.

That said, I think perhaps we could better model our `Server` off of hyper's [Server](https://docs.rs/hyper/latest/hyper/server/struct.Server.html); theirs is _already_ a bounds and listening server upon construction thus there doesn't need to be a `start` (or `spawn`) method.

That is, you contruct/build a server, it internally binds based on its configuration; the binding can be inspected via a method; in the case of hype there is only a [single bound address](https://docs.rs/hyper/latest/hyper/server/struct.Server.html#method.local_addr) for the server, but our implementation might have multiple bound addresses as there might be multiple hyper servers internally (e.g. API, monitoring, etc).

Finally, it implements `Future`, so that you simply await it for it to run to graceful completion.

PR #114, opened 2023-05-15
Refactoring of REST API, client, and server.

### LINKED ISSUE #99: Recommend API routes have `v1` prefix

### LINKED ISSUE #104: Cleanup: request types in the API crate should borrow their data.
Currently, the request types in the `warg-api` crate own their data, which is unnecessary given that they are immediately serialized into JSON.

This results in a few places where we have unnecessary clones.

Instead, the request types should borrow the data.

peterhuene:
This PR includes a sizable refactoring to the client and server to conform to the proposed spec in #101.

It includes improvements to the REST API that enable the following features:

* Fetching logs now supports paging for logs with lots of records.
* The server now informs the client of what content it should upload to start processing a newly published record, rather than the client preemptively sending all the content to the server and giving it URLs.
* The client now downloads a package's content by requesting the record and downloading the content via the sources associated with the record, rather than having explicit knowledge of where the content is located.

Content uploads are now tied to records that are being published; thus the content service (which allowed completely arbitrary uploads) has been removed and the relevant functionality folded into the package service.

Additionally, the REST API was moved under a `v1` path and separated out in both the client and server for future versions.

CORS was also enabled for the API to allow for it to be used from browsers.

Note: a new DB migration was not created for this work as no one is really using the database feature just yet; in the future when folks are actually using the database, a migration would be required.

Closes #104.
Closes #99.


crates/server/src/bin/warg-server.rs
```diff
@@ -24,40 +24,36 @@ struct Args {
...
     database_url_file: Option<PathBuf>,
 
     /// The operator key.
     ///
-    /// Prefer using warg-operator-key-file, or environment variable variation.
-    #[arg(long, env = "WARG_OPERATOR_KEY")]
-    warg_operator_key: Option<String>,
+    /// Prefer using `operator-key-file`, or environment variable variation.
+    #[arg(long, env = "OPERATOR_KEY")]
+    operator_key: Option<String>,
```
Kylebrown9, line 52: Try to make this `Option<PrivateKey>` so that we never treat the secret as a string in Rust code and e.g. debug print it.

crates/server/src/datastore/postgres/schema.rs
```diff
@@ -24,6 +20,17 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    contents (id) {
```
Kylebrown9, line 24: rename to local_contents?

crates/server/src/bin/warg-server.rs
peterhuene: Unfortunately clap requires `Clone` on option fields, which `PrivateKey` doesn't implement intentionally.

Going to leave as-is now with the hope that this option will be removed in the near-term in favor of pulling the key from a key ring.

Kylebrown9 (approved):

tests/client.rs
```diff
@@ -0,0 +1,88 @@
...
+        .await?;
+
+    // Initialize a new package log with releases of the same content
+    // As the number of packages exceeds the limit for fetching in a single request,
+    // this will guarantee that the later call to `upsert` fetches the package log incrementally
+    let mut entries = Vec::with_capacity(PACKAGE_COUNT + 1);
+    entries.push(PublishEntry::Init);
+
+    for i in 1..=PACKAGE_COUNT {
+        entries.push(PublishEntry::Release {
```
Kylebrown9, line 44: Making a large number of entries in one update still only makes one record. Does this test what we're trying to test?

peterhuene: erm whoops! In my attempt to cut down on the calls on publish, I completely invalidate the premise of the test 🤦 .

Let me also run it before the fix commit to ensure it is actually catching the bug.

PR #115, opened 2023-05-15
Fix issues with the recent keyring integration.

peterhuene:
This PR fixes one bug in the keyring integration (#108) and updates the documentation for client configuration.


fibonacci1729 (approved):
Ah, thank you!

PR #117, opened 2023-05-17
chore: update installation & getting started instructions

cardoso:
I think this is outdated, probably there was only a single binary when it was written.

I suppose the cli and server don't need to be installed in the same machine, but this is a quick start right? :)


Kylebrown9:
There were PRs in flight that changed what you were documenting and the docs as well, so I deferred reviewing this till now.

The big question imo is what a "quick start" even means/will mean for the registry since before long there will be live registry servers you can use instead of installing/running your own. For now, this is fine and we can always change it later.

Kylebrown9 (approved):

calvinrp:
The install instructions have since changed. Closing this PR.

PR #121, opened 2023-05-25
Implement a WebAssembly content policy in the server.

peterhuene:
This PR implements a content policy in the server that incrementally validates that content uploaded to the server is valid WebAssembly.

Currently the policy is not exposed via the CLI, but the integration tests use the policy to expect that all content should be WebAssembly.

We may want to expose enabling some policies via the CLI in the future, but for now this should allow for consumers of the `warg-server` crate to enable policies for custom server implementations.


crates/api/src/v1/package.rs
```diff
@@ -120,6 +120,9 @@ pub enum PackageError {
     /// The operation was not supported by the registry.
     #[error("the requested operation is not supported: {0}")]
     NotSupported(String),
+    /// The provided content was rejected by server policy.
+    #[error("the content was rejected by server policy: {0}")]
+    ContentPolicyRejection(String),
```
Kylebrown9, line 125: I wonder if we need to make it specifically a `ContentPolicyRejection` instead of just a `Rejection` with reason. I think there are going to be several kinds of policies and I think in many cases e.g. "content is malformed" some manual effort is going to be needed anyway to solve the problem and we don't need to categorize them in detail.

peterhuene: Agreed, I'll make this more generic.

tests/server.rs
```diff
@@ -132,6 +132,54 @@ async fn validate_wit_package(config: &Config, client: &FileSystemClient) -> Res
     Ok(())
 }
 
+async fn validate_content_policy(client: &FileSystemClient) -> Result<()> {
```
Kylebrown9, line 135: This feels like it should just be part of the test since it performs a specific set of operations and validates them. Does it actually define any generic/shared logic?

Kylebrown9:
I think there will need to be more work in the future to generalize kinds of policies and where they hook in, but that doesn't need to be part of this PR.

LGTM other than one remaining comment.

tests/server.rs
peterhuene: It's shared between the in-memory and postgres tests.

Kylebrown9: From what I see, `it_rejects_non_wasm_content` is the only test that calls this function.

peterhuene: Also called from function `test` in `tests/postgres/mod.rs` (one giant test case that tests against a single postgres server for reasons outlined in the comments).

Kylebrown9 (approved):

Issue #66, opened 2023-01-26
Implement proof for map not containing a value for a key.


Kylebrown9:
The registry can lie to clients that there are no entries for a package because there is no way to prove that at a Map Checkpoint there was no value associated with a given key.

In the REST API, requesting to prove that a log was empty can be represented as
> POST /proofs/inclusion
```json
{
    "checkpoint": ...,
    "logs": [{ "log": "sha256:...", "record": null }]
}
```

A proof that a key was empty is a walk from the first non-empty ancestor of the leaf for that key up to the root.

PR #123, opened 2023-05-25
Sparse Merkle Tree

macovedj:
This PR will change how the sparse merkle tree is optimized.  Currently it implements certain optimizations that are beneficial in some regard, but there are trade-offs.  [Other literature](https://github.com/google/trillian/blob/master/docs/papers/RevocationTransparency.pdf) references this other optimization, which after discussion, we find has benefits that we would like.  Specifically, it will come in handy with respect to the implementation of [this issue](https://github.com/bytecodealliance/registry/issues/66).  [These notes](https://hackmd.io/6RI9ZyGqR1ykRITggTb-pg) may be helpful for understanding the refactor.


crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +41,32 @@ impl FromStr for HashAlgorithm {
...
 
-pub trait SupportedDigest: Digest + private::Sealed {
+static EMPTY_HASH_CACHE: Lazy<HashMap<u8, Hash<Sha256>>> = Lazy::new(|| {
+    let mut m: HashMap<u8, Hash<Sha256>> = HashMap::new();
+    fn empty_hash<D: SupportedDigest>(m: &mut HashMap<u8, Hash<D>>, n: u8) -> Hash<D> {
+        let hash: Hash<D> = if n == 0 {
+            Hash::of("")
+        } else {
+            let last_hash = empty_hash(m, n - 1);
+            Hash::of(Hash::<Sha256>::of((&last_hash, &last_hash)))
```
Kylebrown9, line 51: The logic for branch and leaf hashes should be updated in [`map.rs`](https://github.com/bytecodealliance/registry/blob/main/crates/transparency/src/map/map.rs#L241) and the empty value hash should be added there too.

crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +41,32 @@ impl FromStr for HashAlgorithm {
     }
 }
 
-pub trait SupportedDigest: Digest + private::Sealed {
+static EMPTY_HASH_CACHE: Lazy<HashMap<u8, Hash<Sha256>>> = Lazy::new(|| {
+    let mut m: HashMap<u8, Hash<Sha256>> = HashMap::new();
```
Kylebrown9, line 45: There's no need for a hashmap since this will always be built up sequentially. Use a `Vec` instead and simply build it up iteratively with roughly `hashes.push(branch_hash(last, last))`.

crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +41,32 @@ impl FromStr for HashAlgorithm {
...
+        m.insert(n, hash.clone());
+        hash
+    }
+    empty_hash(&mut m, 255);
+    m
+});
+
+pub trait SupportedDigest: Digest + private::Sealed + Sized {
     const ALGORITHM: HashAlgorithm;
+    fn empty_hash_cache(n: &u8) -> Hash<Self>;
```
Kylebrown9, line 62: Change to `empty_tree_hash(height: u8)`

crates/transparency/src/map/fork.rs
```diff
@@ -31,14 +31,14 @@ impl<D: SupportedDigest> Clone for Fork<D> {
 impl<D: SupportedDigest> Default for Fork<D> {
```
Kylebrown9, line 31: I'm not sure we should have a default anymore. I can't think of a use for it and it'd be easy to use incorrectly.

crates/transparency/src/map/map.rs
```diff
@@ -261,14 +261,9 @@ where
...
 where
     D: SupportedDigest,
 {
-    match (lhs, rhs) {
-        (Some(left), Some(right)) => Hash::of(&(0b11, left, right)),
-        (Some(left), None) => Hash::of(&(0b10, left)),
-        (None, Some(right)) => Hash::of(&(0b01, right)),
-        (None, None) => Hash::of(0b00u8),
-    }
+    Hash::of(&(lhs, rhs))
```
Kylebrown9, line 268: Lann pointed out earlier that we should still use a prefix for leaf and branch so that collisions aren't possible even if the client does something a little wrong.
```suggestion
    Hash::of(&(0b1, lhs, rhs))
```

crates/transparency/src/map/proof.rs
```diff
@@ -34,23 +34,23 @@ where
     V: VisitBytes,
 {
     value: PhantomData<V>,
-    peers: Vec<Option<Hash<D>>>,
+    peers: Vec<Hash<D>>,
```
Kylebrown9, line 37: Ironically, this might want to become `Option<..>` again but for a different reason. If we ever mark something as `None` that can mean that the client knows it's an empty subtree and can substitute `empty_tree_hash(height)` for the value.

crates/transparency/src/map/fork.rs
macovedj: Can push this up with some other minor changes to address other comments.  Before I do, there are two places where `Fork::Default`  currently gets called.  Not sure that I see a way around making `Fork` fields public if we wanted to do this.  Not sure how we value the tradeoffs.

Kylebrown9: In place of a `Default` which I think is ambiguous, you can add a specifically named constructor that creates the needed `Fork`. I'd do this for other defaults too. We should only have a `Default` if that has an actual meaning for a type.

crates/transparency/src/map/mod.rs
```diff
@@ -108,14 +108,12 @@ mod test {
 
     #[test]
     fn prove() {
-        fn check<D: SupportedDigest, K: VisitBytes + Clone, V: VisitBytes>(
+        fn check<D: SupportedDigest + std::fmt::Debug, K: VisitBytes + Clone, V: VisitBytes>(
```
Kylebrown9, line 111: Confused why this is necessary when `Hash<D> where D: SupportedDigest` has a [`fmt::Debug` implementation](https://github.com/bytecodealliance/registry/blob/main/crates/crypto/src/hash/static.rs#L98).

crates/transparency/src/map/fork.rs
```diff
@@ -7,14 +7,18 @@ use warg_crypto::hash::{Hash, SupportedDigest};
 use super::{link::Link, map::hash_branch, path::Side};
 
 pub struct Fork<D: SupportedDigest> {
-    left: Option<Arc<Link<D>>>,
-    right: Option<Arc<Link<D>>>,
+    pub left: Arc<Link<D>>,
```
Kylebrown9, line 10: why `pub`?

crates/server/src/services/core.rs
```diff
@@ -203,9 +203,9 @@ impl CoreService {
         while let Some(res) = initial.next().await {
             let InitialLeaf { leaf, checkpoint } = res?;
             data.log.push(&leaf);
-
+            let new_key = leaf.log_id.0.clone().try_into().unwrap();
```
peterhuene, line 206: This doesn't appear to be any different from the existing `leaf.log_id.clone()`, or am I mistaken?

It appears to just be invoking `From<AnyHash> for LogId`, which we're cloning; cloning the `LogId` is just clearer to me.

crates/transparency/src/map/path.rs
```diff
@@ -51,42 +49,73 @@ impl<D: SupportedDigest> Path<D> {
...
+    fn clone(&self) -> Self {
+        Self {
+            all: self.all.clone(),
+            index: self.index,
+        }
+    }
+}
+
+impl<D: SupportedDigest> ReversePath<D> {
+    pub(crate) fn new<K: VisitBytes>(key: K) -> Self {
```
peterhuene, line 87: It doesn't look like we need to take `key` by value here, as we're only storing the hash.

```suggestion
    pub(crate) fn new<K: VisitBytes>(key: &K) -> Self {
```

crates/transparency/src/map/path.rs
```diff
@@ -28,15 +26,15 @@ impl Side {
...
 
-        Self {
-            lhs: 0,
-            rhs: all.len() * 8,
-            all,
-        }
+        Self { key, index: 0, all }
+    }
+
+    pub fn key(&self) -> K {
```
peterhuene, line 36: A key (i.e. ` LogId`) might have a clone that allocates, so can we return `&K` here so that the caller gets to decide if a clone is necessary?

crates/transparency/src/map/singleton.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self {
+            key,
+            value,
+            height,
+            side,
+        }
+    }
+
+    pub fn key(&self) -> K {
+        self.key.clone()
```
peterhuene, line 30: Same comment as above as to why I think this should return `&K`.

crates/transparency/src/map/node.rs
```diff
@@ -1,51 +1,78 @@
...
-                Some(proof)
+                let proof = fork[idx].as_ref().node().prove(path);
+                match proof {
+                    Some(mut p) => {
+                        let peer = fork[idx.opposite()].as_ref().hash();
+                        p.push(Some(peer.clone()));
+                        Some(p)
+                    }
+                    None => None,
+                }
```
peterhuene, line 75:
```suggestion
                let mut proof = fork[idx].as_ref().node().prove(path)?;
                let peer = fork[idx.opposite()].as_ref().hash();
                proof.push(Some(peer.clone()));
                Some(proof)
```

crates/client/src/api.rs
```diff
@@ -401,12 +401,12 @@ impl Client {
...
         }
 
-        let map_proof_bundle: MapProofBundle<Sha256, MapLeaf> =
+        let map_proof_bundle: MapProofBundle<Sha256, LogId, MapLeaf> =
             MapProofBundle::decode(response.map.as_slice())?;
         let map_inclusions = map_proof_bundle.unbundle();
         for (leaf, proof) in leafs.iter().zip(map_inclusions.iter()) {
             let found = proof.evaluate(
-                &leaf.log_id,
+                leaf.log_id.clone(),
```
peterhuene, line 409: If `ReversePath` is made to take `&K` in its constructor, I think we can back out this change.

crates/server/src/services/core.rs
```diff
@@ -248,12 +248,15 @@ impl CoreService {
 
         store.validate_operator_record(&log_id, &record_id).await?;
 
-        let leaf = LogLeaf { log_id, record_id };
+        let leaf = LogLeaf {
+            log_id: log_id.clone(),
```
peterhuene, line 252: I think this clone is unnecessary (i.e. we can back out this change along with the change to inserting into the map below).

crates/transparency/src/map/proof.rs
```diff
@@ -28,72 +29,75 @@ use super::{
...
-
-    #[cfg(test)]
-    pub(crate) fn len(&self) -> usize {
-        self.peers.len()
+        self.peers.push(peer);
     }
 
     /// Computes the root obtained by evaluating this inclusion proof with the given leaf
-    pub fn evaluate<K: ?Sized + VisitBytes>(&self, key: &K, value: &V) -> Hash<D> {
+    pub fn evaluate(&self, key: K, value: &V) -> Hash<D> {
```
peterhuene, line 63: We can still take `key` as `&K` if `ReversePath::new` takes `&K` (which it can).

crates/server/src/services/transparency/map.rs
```diff
@@ -49,7 +49,8 @@ pub fn spawn(input: Input) -> Output {
                 summary = log_summary_rx.recv() => {
                     if let Some(summary) = summary {
                         let leaf = summary.leaf;
-                        map = map.insert(leaf.log_id.clone(), MapLeaf { record_id: leaf.record_id.clone() });
+                        let hash = leaf.log_id.0.clone().try_into().unwrap();
```
peterhuene, line 52: Likewise, I don't think this conversion is needed and we can continue to just insert `leaf.log_id.clone()`, unless I'm mistaken?

peterhuene (commented):
Just some (mostly nit) comments I noticed while trying to determine the source of the CI failure.

I did verify that the server restart with the postgres datastore does appear to reconstruct the transparency service data structures as expected.

I'm not sure why it's only failing to prove after a restart of the server, though.

crates/transparency/src/map/map.rs
```diff
@@ -224,51 +185,18 @@ where
...
-/// H(0b10 || <left>)
-///
-/// // Right child present:
-/// H(0b01 || <right>)
-///
-/// // Empty tree:
-/// H(0b00)
-/// ```
-pub(crate) fn hash_branch<D>(lhs: Option<Hash<D>>, rhs: Option<Hash<D>>) -> Hash<D>
+pub(crate) fn hash_branch<D>(lhs: Hash<D>, rhs: Hash<D>) -> Hash<D>
```
peterhuene, line 197: Can we pass `lhs` and `rhs` by reference? Doesn't seem like this needs to be by-value.

crates/server/src/services/data/map.rs
```diff
@@ -23,7 +23,7 @@ pub struct Output {
 
 #[derive(Default)]
 pub struct MapData {
-    map_index: HashMap<Hash<Sha256>, VerifiableMap>,
+    pub map_index: HashMap<Hash<Sha256>, VerifiableMap>,
```
peterhuene, line 26: This doesn't appear to need `pub`?

crates/transparency/benches/map.rs
```diff
@@ -2,7 +2,7 @@ use std::{iter::repeat_with, time::Duration};
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use rand::Rng;
-use warg_crypto::hash::Sha256;
+use warg_crypto::hash::{AnyHash, Hash, Sha256};
```
peterhuene, line 5: I think this change can be backed out.

crates/server/src/services/data/map.rs
```diff
@@ -35,28 +35,25 @@ impl MapData {
...
         let map = self
             .map_index
             .get(root)
             .ok_or_else(|| DataServiceError::RootNotFound(root.clone()))?;
-
         let mut proofs = Vec::new();
         for LogLeaf { log_id, record_id } in leaves {
-            let proof = map
-                .prove(log_id)
-                .ok_or_else(|| DataServiceError::PackageNotIncluded(log_id.clone()))?;
```
peterhuene, line 48: I think we'd still want this early return of `DataServiceError::PackageNotIncluded` here as the log id is coming from the request and may not be part of the map, so I'd expect an error rather than the `unwrap` call below panicking.

crates/transparency/benches/map.rs
peterhuene: Looks like we are left with a change to add brackets around `Sha256`, can probably just revert the whole file.

crates/protocol/src/registry.rs
```diff
@@ -166,18 +166,18 @@ impl<'de> Deserialize<'de> for PackageId {
...
         let prefix: &[u8] = b"WARG-OPERATOR-LOG-ID-V0".as_slice();
-        let hash: Hash<D> = Hash::of(prefix);
+        let hash: Hash<D> = Hash::of(Hash::<D>::of(prefix));
         Self(hash.into())
     }
 
     pub fn package_log<D: SupportedDigest>(id: &PackageId) -> Self {
         let prefix: &[u8] = b"WARG-PACKAGE-LOG-ID-V0:".as_slice();
-        let hash: Hash<D> = Hash::of((prefix, id));
+        let hash: Hash<D> = Hash::of(Hash::<D>::of((prefix, id)));
```
peterhuene, line 180: Before we merge this, can we document why this double hash is necessary here (and likewise for the operator log id)?

At first glance, a `LogId` functions simply as an identifier, meaning as long as the same identifier is derived from the same package identifier (and internal prefix), then I would expect a single hash to suffice.

If there is some expected way this interacts with the transparency data structures, it should be called out explicitly.

crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +40,32 @@ impl FromStr for HashAlgorithm {
...
+            Hash::of(Hash::<Sha256>::of((0b1, &last_hash, &last_hash)))
+        };
+        v.push(hash.clone());
+        hash
+    }
+    empty_tree_hash(&mut v, 256);
+    v
+});
+
+pub trait SupportedDigest: Digest + private::Sealed + Sized + std::fmt::Debug {
```
Kylebrown9, line 59: why is there a `std::fmt::Debug` bound here?
it shouldn't be necessary since `Hash<D> where D: SupportedDigest` impls `Debug`.

crates/protocol/src/registry.rs
Kylebrown9: I agree with Peter that this should not need to be hashed twice.

A package's `LogId` should be a simple hash of the prefix string and the name.

crates/transparency/src/map/fork.rs
```diff
@@ -1,44 +1,52 @@
...
-        Self {
-            left: None,
-            right: None,
-        }
-    }
-}
-
-impl<D: SupportedDigest> Index<Side> for Fork<D> {
-    type Output = Option<Arc<Link<D>>>;
+impl<D: SupportedDigest, K: std::fmt::Debug + VisitBytes + Clone + PartialEq> Index<Side>
```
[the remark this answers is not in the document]
Kylebrown9: Or, if we do need a crapload of bounds for keys, we may want a `MapKey` trait that bundles all those bounds/requirements up, providing a shorthand and making it more obvious to users what's needed.

crates/transparency/src/map/map.rs
```diff
@@ -41,48 +41,13 @@ use super::proof::Proof;
 /// to have `2^(n+1) - 1` entries (far too many!), the tree is sparse and
 /// only creates nodes as necessary to represent the contents in the tree.
 ///
-/// ## Hashing Strategy
```
Kylebrown9, line 44: Please update documentation instead of removing it.

crates/transparency/src/map/mod.rs
```diff
@@ -108,30 +109,34 @@ mod test {
...
-            peers: usize,
         ) {
-            let proof = tree.prove(&key).unwrap();
-            assert_eq!(proof.len(), peers);
+            let proof = tree.prove(key.clone()).unwrap();
             assert_eq!(tree.root().clone(), proof.evaluate(&key, &value));
         }
 
         let first = Map::<Sha256, &'static str, &'static str>::default();
-        assert!(first.prove(&"foo").is_none());
```
Kylebrown9, line 123: This should be able to take a reference.

crates/transparency/src/map/node.rs
```diff
@@ -1,50 +1,71 @@
...
             Self::Fork(node) => Self::Fork(node.clone()),
+            Self::Singleton(s) => Self::Singleton(s.clone()),
+            Self::Empty(height) => Self::Empty(*height),
         }
     }
 }
 
-impl<D: SupportedDigest> Default for Node<D> {
+impl<D: SupportedDigest, K: Debug + VisitBytes + Clone + PartialEq> Default for Node<D, K> {
     fn default() -> Self {
```
Kylebrown9, line 34: Why is this the default and what does it mean?

crates/transparency/src/map/node.rs
```diff
@@ -1,51 +1,78 @@
...
+                if singleton.key() == path.key() {
+                    Some(Proof::new(Vec::new()))
+                } else {
+                    None
+                }
+            }
+            (Some(_), Self::Empty(_)) => {
+                let mut proof: Proof<D, K, V> = self.prove(path)?;
+                proof.push(None);
+                None
```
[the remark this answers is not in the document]
Kylebrown9: I'm also not sure why we're calling `prove` again. I believe this represents the case where you're trying to prove something exists in an empty subtree, which should be able to eagerly return `None`.

crates/transparency/src/map/node.rs
```diff
@@ -59,27 +80,134 @@ impl<D: SupportedDigest> Node<D> {
...
-                    let node = fork[index]
-                        .as_ref()
-                        .map(|link| link.node().clone())
-                        .unwrap_or_default();
-
-                    // Replace its value recursively.
-                    let (node, new) = node.insert(path, leaf);
-                    fork[index] = Some(Arc::new(Link::new(node)));
-                    (Node::Fork(fork), new)
+                    let node = fork[index].as_ref().node();
```
Kylebrown9, line 118: Why doesn't this just recurse into `fork[index].as_ref().node().insert(...)`?

crates/transparency/src/map/node.rs
```diff
@@ -59,27 +80,134 @@ impl<D: SupportedDigest> Node<D> {
...
+                    if singleton.key() == &key {
+                        let new_singleton = Singleton::new(key, value, 256 - path.index(), index);
+                        (Node::Singleton(new_singleton), false)
+                    } else if singleton.side != index {
+                        let node =
+                            Node::Singleton(Singleton::new(key, value, 256 - path.index(), index));
+                        let original = Node::Singleton(Singleton::new(
+                            singleton.key,
+                            singleton.value,
+                            256 - path.index(),
```
Kylebrown9, line 166: Create a method `path::height()` that does this `256 - path.index()` logic.

crates/transparency/src/map/path.rs
```diff
@@ -1,14 +1,12 @@
-use core::iter::FusedIterator;
-
 use warg_crypto::{
     hash::{Hash, SupportedDigest},
     VisitBytes,
 };
 
-pub struct Path<D: SupportedDigest> {
+pub struct Path<D: SupportedDigest, K: VisitBytes + Clone> {
+    key: K,
```
Kylebrown9, line 7: Why not borrow?

crates/transparency/src/map/singleton.rs
```diff
@@ -0,0 +1,58 @@
...
+    map::hash_branch,
+    path::{ReversePath, Side},
+};
+
+#[derive(Debug)]
+pub struct Singleton<D: SupportedDigest, K: VisitBytes + Clone> {
+    pub key: K,
+    pub value: Hash<D>,
+    pub height: usize,
+    pub side: Side,
```
Kylebrown9, line 16: Why is `size: Side` needed?

crates/transparency/src/map/map.rs
Kylebrown9: Bump

crates/transparency/src/map/node.rs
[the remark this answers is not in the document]
Kylebrown9: Bump

crates/transparency/src/map/node.rs
```diff
@@ -59,27 +74,59 @@ impl<D: SupportedDigest> Node<D> {
     /// Returns:
     ///   * the new node that must replace the current node.
     ///   * whether or not this is a new entry in the map.
-    pub fn insert(&self, path: &mut Path<D>, leaf: Hash<D>) -> (Self, bool) {
+    pub fn insert(&self, path: &mut Path<'_, D>, value: Hash<D>) -> (Self, bool) {
```
Kylebrown9, line 77: Rust might let you remove this lifetime on `Path` since it's the elided lifetime `'_`.

crates/transparency/src/map/singleton.rs
Kylebrown9: Bump.

crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +40,45 @@ impl FromStr for HashAlgorithm {
...
+
+// If updating this function, also update `hash_branch` in transparency map
+pub(crate) fn hash_branch<D>(lhs: &Hash<D>, rhs: &Hash<D>) -> Hash<D>
+where
+    D: SupportedDigest,
+{
+    Hash::of((0b1, lhs, rhs))
+}
+
+pub trait SupportedDigest: Digest + private::Sealed + Sized + 'static {
```
Kylebrown9, line 72: Why is `Sized` here?

crates/transparency/src/map/node.rs
macovedj: Can remove it, but `cargo check` still warns

crates/crypto/src/hash/mod.rs
macovedj: Returning `Hash<Self>` requires it

crates/transparency/src/map/node.rs
Kylebrown9: weird, wouldn't guess that it'd be legal but a warning. no worries

crates/transparency/src/map/map.rs
```diff
@@ -43,44 +43,20 @@ use super::proof::Proof;
...
-///
-/// // Right child present:
-/// H(0b01 || <right>)
-///
-/// // Empty tree:
-/// H(0b00)
-/// ```
+/// #Branch nodes
+/// The value at a node is the hash of its two children, prepended with a
+/// 1 bit (e.g. 0b1 || Hash(left_val) || Hash(right_val))
```
Kylebrown9, line 54: Rephrase branch hash strategy
```suggestion
/// ### Branch Nodes
///
/// Branch node hashes are calculated using the one-byte `1` prefix to prevent collisions with branch hashes.
/// ```hash
/// hash(Branch(left, right)) = hash(0x01 || hash(<left>) || hash(<right>))
/// ```
```

crates/transparency/src/map/map.rs
```diff
@@ -43,44 +43,20 @@ use super::proof::Proof;
 ///
 /// ## Hashing Strategy
 ///
-/// Hashes for leaf and branch nodes are calculated using the following input
-/// pattern:
+/// #Leaf nodes
+/// Initially, all leaves have not been inserted into, so their hash value
+/// simply the hash of an empty string.  When a leaf has a value inserted, we
+/// insert the hash of the value prepended with a 0 bit (e.g. 0b0 || Hash(value)
+/// at that node
```
Kylebrown9, line 50: Rephrase leaf hash strategy and make a change to how "empty" leaves are interpreted.
```suggestion
/// ### Leaf Nodes
///
/// Leaf node hashes are calculated using the one-byte `0` prefix to prevent collisions with branch hashes.
///
/// ```hash
/// H(0x00 || <value>)
/// ```
///
/// For leaf nodes that semantically "contain no value", the `<value>` provided is empty i.e. the zero-length byte sequence.
///
```

crates/transparency/src/map/map.rs
```diff
@@ -224,51 +197,50 @@ where
...
+// If updating this function, also update `hash_branch` in crypto crate
+/// Compute the hash for an empty leaf using a given Digest algorithm.
+#[allow(dead_code)]
+pub(crate) fn hash_empty<D: SupportedDigest>() -> Hash<D> {
+    Hash::of(())
+}
+
+// No associated function exists in crypto crate today, but in the event that one exists
+// update this function if updating the other
+/// Compute the hash for a leaf with a given value using a given Digest algorithm
```
Kylebrown9, line 219:
```suggestion
// No associated function exists in crypto crate today, but in the event that one exists
// update this function if updating the other
/// Hashes a leaf node. See [Map] docs for more detail.
```

crates/transparency/src/map/map.rs
```diff
@@ -224,51 +197,50 @@ where
...
-///
-/// // Right child present:
-/// H(0b01 || <right>)
-///
-/// // Empty tree:
-/// H(0b00)
-/// ```
-pub(crate) fn hash_branch<D>(lhs: Option<Hash<D>>, rhs: Option<Hash<D>>) -> Hash<D>
+// If updating this function, also update `hash_branch` in crypto crate
+/// Compute the hash for a branch node using a given Digest algorithm
```
Kylebrown9, line 229:
```suggestion
/// Hashes a branch node. See [Map] docs for more detail.
```

crates/transparency/src/map/map.rs
```diff
@@ -224,51 +197,50 @@ where
...
-/// Hashes a leaf node
-/// ```hash
-/// H(0xff || H(<key>) || <value>)
-/// ```
-pub(crate) fn hash_leaf<D, K, V>(key: K, value: V) -> Hash<D>
+// If updating this function, also update `hash_branch` in crypto crate
+/// Compute the hash for an empty leaf using a given Digest algorithm.
+#[allow(dead_code)]
+pub(crate) fn hash_empty<D: SupportedDigest>() -> Hash<D> {
+    Hash::of(())
```
Kylebrown9, line 214:
```suggestion
    leaf_hash(())
```

crates/transparency/src/map/map.rs
```diff
@@ -224,51 +197,50 @@ where
...
         here
     }
 }
 
-/// Hashes a leaf node
-/// ```hash
-/// H(0xff || H(<key>) || <value>)
-/// ```
-pub(crate) fn hash_leaf<D, K, V>(key: K, value: V) -> Hash<D>
+// If updating this function, also update `hash_branch` in crypto crate
```
Kylebrown9, line 210:
```suggestion
// If updating this function, also update `hash_empty` in crypto crate
```

crates/crypto/src/hash/mod.rs
Kylebrown9: I believe we can add `!Sized` on `Hash` or somewhere else so that this isn't necessary, but it's not worth worrying about it here/now.

crates/crypto/src/hash/mod.rs
```diff
@@ -39,12 +40,45 @@ impl FromStr for HashAlgorithm {
...
+        };
+        v.push(hash.clone());
+        hash
+    }
+    empty_tree_hash(&mut v, 256);
+    v
+});
+
+// If updating this function, also update `hash_empty` in transparency map
+pub(crate) fn hash_empty<D: SupportedDigest>() -> Hash<D> {
```
Kylebrown9, line 60: Copy `hash_leaf` over here and update this `hash_empty` to call it like the suggestion in the other location.

Kylebrown9 (approved):
Thanks for all the work you've done to bring this together!

PR #125, opened 2023-05-26
Update all dependencies to latest.

peterhuene:
This PR updates all of the dependencies for the workspace to the latest versions.

This includes support for the latest binary format of the component model for the content policy.

### DIFF
--- a/tests/postgres/mod.rs
+++ b/tests/postgres/mod.rs
@@ -12,9 +12,9 @@ fn data_store() -> Result<Box<dyn DataStore>> {
     )?))
 }
 
-async fn run<F>(root: &Path, callback: impl FnOnce(Config) -> F) -> Result<()>
+async fn run<F, T>(root: &Path, callback: impl FnOnce(Config) -> F) -> Result<T>
 where
-    F: Future<Output = Result<()>>,
+    F: Future<Output = Result<T>>,
 {
     let (_server, config) = super::support::spawn_server(root, Some(data_store()?)).await?;
     callback(config).await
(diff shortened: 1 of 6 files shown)


tests/server.rs
```diff
@@ -57,7 +57,7 @@ async fn validate_component_package(config: &Config, client: &FileSystemClient)
         .context("failed to resolve package")?;
     assert_eq!(
         download.digest.to_string(),
-        "sha256:396bf81fe30c615180c31fc3ba964396241af472ace265f55609a3fcf12140f2"
+        "sha256:24eaf3439a60c96764c4f2c92a0a81b52bd4fe5d5b91f090184a0f6486b1bf29"
```
Kylebrown9, line 60: Thoughts on deriving this hash automatically from the content so that it doesn't have to be changed when the content does?

peterhuene: I can add that.

tests/postgres/mod.rs
```diff
@@ -46,24 +46,27 @@ async fn test() -> Result<()> {
     .await?;
 
     // Publish and validate a component package
-    run(&root, |config| async move {
-        let client = super::create_client(&config)?;
-        super::publish_component_package(&client).await?;
-        super::validate_component_package(&config, &client).await?;
+    let digest = run(&root, |config| async move {
+        let client: warg_client::Client<
```
peterhuene, line 50: Weird, I must have accidentally hit a rust-analyzer tooltip to explicitly qualify this, which isn't necessary. I'll remove.

Kylebrown9 (approved):

PR #126, opened 2023-05-31
Renamed DynHash to AnyHash

calvinrp:
Avoids the possible confusion associating with the `dyn` Rust keyword.


peterhuene (approved):
Thanks for fixing this (it's a much better name).

Issue #94, opened 2023-04-13
Refactor some names


Kylebrown9:
We've come up with some clearer/better names for types the more we use them. Now we just need to swap them out.

* `Validator` in e.g. `PackageValidator` -> `State`
  * While you can validate using these, the struct itself really represents the state of the validator.
*  `DynHash`/`Hash<D>` -> `AnyHash`/`Hash<D>`
   * The "static vs. dynamic" naming here is accurate in some sense, but is confusing especially with the use of `dyn` as a keyword for things. The real distinction is that `AnyHash` can hold a hash of any allowed algorithm type whereas `Hash<D>` identifies a singular algorithm in its type.

peterhuene:
Second half of this issue is fixed with #126.

calvinrp:
> For renaming `Validator`, how about `LogState`?

Sounds fine to me

PR #128, opened 2023-06-03
Renamed Validator implementation for Operator and Package records

calvinrp:
Although it is a validator (and implements `Validator` trait), it is used to reflect the state of the log. Renamed to `OperatorState` and `PackageState`. See issue: https://github.com/bytecodealliance/registry/issues/94


calvinrp:
There are ~7 files that import both into the same namespace. I'm on board with those suggestions.

Kylebrown9:
I guess for change 2, the logical place might be to just make them all available in the `protocol` crate root. I don't know that we have a better place and creating some new module e.g. `mod log { pub use ... }` in the root or adding a new level to the hierarchy seems a bit silly.

crates/protocol/src/package/mod.rs
```diff
@@ -8,7 +8,7 @@ mod model;
 mod validate;
 
 pub use model::{PackageEntry, PackageRecord, Permission};
-pub use validate::{Head, PackageState, Release, ReleaseState, ValidationError};
```
Kylebrown9: There's still an inconsistency here between `PackageState`/`PackageValidationError` and `Release`/`ReleaseState`, which I think is a good focal point for the main question we have to answer.

Which of the following do we want in the codebase?
1. all names get log-name prefixing including `PackageRelease` & `PackageReleaseState`;
2. some names get log-name prefixing, but others like `Release` and `ReleaseState` don't;
3. no names are log-name prefixed and we consistently use `package::` and `operator::` for disambiguation like we do today.

I'm not a fan of the inconsistency of 2, without trying them both I don't have a huge preference between 1 & 3. What do other people think?

@lann @peterhuene

lann: I'm fine with 1 or 3. I would probably pick 1 for this particular case.

calvinrp: Option 1 pushed.

calvinrp:
Looks like I have some merge issues that still need fixing

calvinrp:
Going to scrap this one and start over

PR #130, opened 2023-06-06
Implement authorization by publishing key id.

peterhuene:
This PR implements authorization of publishing operations by the key id used to sign the record being published.

By default, the server `Config` denies all keys, and keys can be specified per-namespace with the `with_authorized_key` method.

Authorization can also be disabled (allow any key id to claim a package name) with the `with_no_authorization` method.

A consequence of this change is that package names are now restricted to valid package names in a component model ID (e.g. `wasi:http`).

The tests have also been refactored to better share the test implementations between the in-memory and postgres data stores.

In the future, the authorization scheme could be further refined to specific package names rather than entire namespaces.

### DIFF
--- a/crates/api/src/v1/package.rs
+++ b/crates/api/src/v1/package.rs
@@ -112,11 +112,8 @@ pub enum PackageError {
     #[error("the record is not currently sourcing content")]
     RecordNotSourcing,
     /// The publish operation was not authorized by the registry.
-    #[error("registry did not authorize the request to publish: {message}")]
-    Unauthorized {
-        /// The error message.
-        message: String,
-    },
+    #[error("registry did not authorize the request to publish: {0}")]
+    Unauthorized(String),
     /// The operation was not supported by the registry.
     #[error("the requested operation is not supported: {0}")]
     NotSupported(String),
@@ -193,7 +190,7 @@ where
 impl Serialize for PackageError {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match self {
-            Self::Unauthorized { message } => RawError::Unauthorized::<()> {
+            Self::Unauthorized(message) => RawError::Unauthorized::<()> {
                 status: Status::<403>,
                 message: Cow::Borrowed(message),
             }
@@ -239,9 +236,9 @@ impl<'de> Deserialize<'de> for PackageError {
         D: serde::Deserializer<'de>,
     {
         match RawError::<String>::deserialize(deserializer)? {
-            RawError::Unauthorized { status: _, message } => Ok(Self::Unauthorized {
-                message: message.into_owned(),
-            }),
+            RawError::Unauthorized { status: _, message } => {
+                Ok(Self::Unauthorized(message.into_owned()))
+            }
             RawError::NotFound { status: _, ty, id } => match ty {
                 EntityType::Log => Ok(Self::LogNotFound(
                     id.parse::<AnyHash>()
--- a/crates/api/src/v1/package.rs
+++ b/crates/api/src/v1/package.rs
@@ -6,7 +6,7 @@ use std::{borrow::Cow, collections::HashMap};
 use thiserror::Error;
 use warg_crypto::hash::AnyHash;
 use warg_protocol::{
-    registry::{LogId, MapCheckpoint, RecordId},
+    registry::{LogId, MapCheckpoint, PackageId, RecordId},
     ProtoEnvelopeBody, SerdeEnvelope,
 };
 
@@ -25,8 +25,8 @@ pub enum ContentSource {
 #[derive(Serialize, Deserialize)]
 #[serde(rename = "camelCase")]
 pub struct PublishRecordRequest<'a> {
-    /// The name of the package being published.
-    pub name: Cow<'a, str>,
+    /// The id of the package being published.
+    pub id: Cow<'a, PackageId>,
     /// The publish record to add to the package log.
     pub record: Cow<'a, ProtoEnvelopeBody>,
     /// The complete set of content sources for the record.
@@ -111,8 +111,8 @@ pub enum PackageError {
     /// The record is not currently sourcing content.
     #[error("the record is not currently sourcing content")]
     RecordNotSourcing,
-    /// The publish operation was not authorized by the registry.
-    #[error("registry did not authorize the request to publish: {0}")]
+    /// The operation was not authorized by the registry.
+    #[error("unauthorized operation: {0}")]
     Unauthorized(String),
     /// The operation was not supported by the registry.
     #[error("the requested operation is not supported: {0}")]
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -22,7 +22,7 @@ use warg_crypto::{
 };
 use warg_protocol::{
     operator, package,
-    registry::{LogId, LogLeaf, MapCheckpoint, RecordId},
+    registry::{LogId, LogLeaf, MapCheckpoint, PackageId, RecordId},
     ProtoEnvelope, SerdeEnvelope, Version, VersionReq,
 };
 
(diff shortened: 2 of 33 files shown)


crates/server/src/api/mod.rs
```diff
@@ -19,11 +24,19 @@ pub fn create_router(
     temp_dir: PathBuf,
     files_dir: PathBuf,
     content_policy: Option<Arc<dyn ContentPolicy>>,
+    authorized_keys: Option<HashMap<String, HashSet<KeyID>>>,
```
Kylebrown9, line 27: At some point I would like policies to be a `Vec<dyn Policy>`-style construction where this namespace control policy isn't special.

In this flow, the `Policy` trait would either
* have methods for each of the possible things it could have an opinion on,
* would have getters/factories that optionally return sub-policies for specific areas,
* or would actually be an `enum` such that each policy refers to a single area and you just add multiple policies to the list if you need to refer to multiple areas.

crates/server/src/api/v1/package.rs
```diff
@@ -85,6 +95,33 @@ impl Config {
...
+/// Valid package names conform to the component model specification
+/// for identifiers.
+///
+/// A valid component model identifier is the format `<namespace>:<name>`,
+/// where both parts are also kebab-case WIT identifiers.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+struct PackageName<'a> {
+    namespace: &'a str,
+    name: &'a str,
+}
```
Kylebrown9, line 109: I think this belongs in the protocol crate next to `LogId` and we'll want to make `LogId` derivable from this.

crates/server/src/datastore/memory.rs
```diff
@@ -577,4 +578,27 @@ impl DataStore for MemoryDataStore {
             checkpoint,
         })
     }
+
+    async fn verify_package_record_signature(
```
Kylebrown9, line 582: What is this function doing?

peterhuene: We need to verify the signature of the record being published prior to accepting the record for processing; to do that, the data store needs to fetch the package's validation state to determine the public key to verify with.

This prevents people from sending us requests where the `keyId` property is explicitly set to an acceptable (known) key id, as these are not private, and then giving us a record signed with a mismatched private key.

By doing so, it preemptively rejects the record before it ever makes it into the database for later validation; validation would have rejected the record anyway due to the invalid signature.

As authorization is based on these public key ids, we shouldn't allow unauthorized keys to write to the database in any form.

crates/server/src/lib.rs
```diff
@@ -108,6 +138,29 @@ impl Config {
         self.content_policy = Some(Arc::new(policy));
         self
     }
+
+    /// Sets an authorized key for a particular namespace.
+    pub fn with_authorized_key(mut self, namespace: impl Into<String>, key: KeyID) -> Result<Self> {
```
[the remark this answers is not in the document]
peterhuene: I'll take a stab at moving this to a proper "record policy" trait with a particular implementation for authorizing.

crates/api/src/v1/package.rs
```diff
@@ -112,11 +112,8 @@ pub enum PackageError {
     #[error("the record is not currently sourcing content")]
     RecordNotSourcing,
     /// The publish operation was not authorized by the registry.
-    #[error("registry did not authorize the request to publish: {message}")]
-    Unauthorized {
-        /// The error message.
-        message: String,
```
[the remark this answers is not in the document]
peterhuene: I can standardize the error types in the other direction instead (there were other tuple variants and this one was the outlier) and make them all struct variants.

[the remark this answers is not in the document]
Kylebrown9: I guess if it's specifically a convention that "error enums with a single message field are mono-tuples" that makes sense to me. I just don't want to lose names when they can't be inferred so easily.

Speaking of which, it might make sense to document some conventions at some point.

crates/server/src/api/mod.rs
peterhuene: I moved this out of `Config` and into an `AuthorizedKeyPolicy` implementing the logic.

The mechanism is minimal right now, but we can expand on the design in future work.

crates/server/src/lib.rs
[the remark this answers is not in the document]
peterhuene: I moved this to the specific policy and `Config` now only has a method to set the record policy to use.

crates/server/src/lib.rs
```diff
@@ -23,6 +24,28 @@ pub mod services;
 const DEFAULT_BIND_ADDRESS: &str = "127.0.0.1:8090";
 const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(5);
 
+fn is_kebab_case(s: &str) -> bool {
```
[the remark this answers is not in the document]
peterhuene: Agreed, `wit_parser` should probably expose this; right now this (and the "interface id" parsing) is crate internal.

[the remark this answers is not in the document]
peterhuene: Actually, I think we can use `KebabStr` from `wasmparser` for this validation. I'll fix.

crates/client/src/lib.rs
```diff
@@ -246,16 +237,21 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
 
     /// Inserts or updates the logs of the specified packages in client storage to
     /// the latest registry checkpoint.
-    pub async fn upsert(&self, packages: &[&str]) -> Result<(), ClientError> {
+    pub async fn upsert<'a, I>(&self, packages: I) -> Result<(), ClientError>
```
Kylebrown9, line 240: Do we use this in non-slice contexts now?

crates/server/src/policy/record/authorization.rs
```diff
@@ -0,0 +1,63 @@
...
+use anyhow::{bail, Result};
+use std::collections::{HashMap, HashSet};
+use warg_crypto::signing::KeyID;
+use warg_protocol::{package::PackageRecord, registry::PackageId, ProtoEnvelope};
+use wasmparser::names::KebabStr;
+
+/// A policy that ensures a published record is authorized by
+/// the key signing the record.
+#[derive(Default)]
+pub struct AuthorizedKeyPolicy {
```
Kylebrown9, line 11: Thoughts on making this `Serialize + Deserialize` so we can pull it from a config file?

crates/protocol/src/registry.rs
```diff
@@ -76,6 +77,93 @@ impl VisitBytes for LogLeaf {
...
+pub struct PackageId {
+    id: String,
+    colon: usize,
+}
+
+impl PackageId {
+    /// Creates a package identifier from the given string.
+    ///
+    /// Returns an error if the given string is not a valid package identifier.
+    pub fn new(id: impl Into<String>) -> anyhow::Result<Self> {
```
Kylebrown9, line 97: Should this be called `try_new` and/or does it even need to exist outside the parse & serde impls?

crates/server/src/datastore/memory.rs
```diff
@@ -210,7 +211,7 @@ impl DataStore for MemoryDataStore {
     async fn store_package_record(
         &self,
         log_id: &LogId,
-        _name: &str,
+        _package_id: &PackageId,
```
Kylebrown9, line 214: why/how is this unused?

Kylebrown9 (approved):
LGTM

I have a few questions/suggestions, but I'll pre-approve so once addressed you can merge without waiting for me again.

crates/server/src/datastore/memory.rs
peterhuene: The in-memory data store doesn't keep the names around, but the postgres one does (really only so one can look at the actual SQL table and see the package ids are; they aren't read out by the server).

Until we have a query API, it really helps the operator to figure out what logs are present.

crates/server/src/policy/record/authorization.rs
peterhuene: I think it'd be best addressed in a follow-up PR to properly represent a server configuration as a single file that can be deserialized to enable the policies (with their specific configurations).

I purposely punted on that (i.e. there's no way to enable policies in the server CLI) as I would want to have a proper discussion around configuration file format.

crates/client/src/lib.rs
peterhuene: We do in the tests currently. As this method is only iterating on the package ids with no requirement it needs to be a slice, it actually helps consumers like `cargo-component` that are iterating on some other data structure and projecting into it to get the warg package identifiers; this change removes the need for callers to collect into a contiguous collection.

crates/protocol/src/registry.rs
peterhuene: This exists for anyone that wants to move the string into the `PackageId`; in low-level libraries like this, I generally like to give people the option of not cloning if possible.

Happy to call it `try_new` (I feel like the naming around fallible a `new` isn't consistent in a lot of codebases, though; `wasmtime` has a bunch of `new` that return `T` and also `Result<T>`).

lann: Yeah fallible `new` is quite common.

PR #136, opened 2023-06-14
Add `--content-base-url` option to `warg-server`.

peterhuene:
This PR adds a `--content-base-url` to `warg-server` that allows specifying what the expected base URL to use when serving content.

By default, the base URL is derived from the listen address for the server.

For servers that aren't bound to the address that will also be serving the content, the `--content-base-url` option can be used to control the URLs the server returns for fetching content.


calvinrp:
Just to confirm the behavior, does the base URL no longer have a default value? Looks good to me either way.

calvinrp:
> Just to confirm the behavior, does the base URL no longer have a default value? Looks good to me either way. From your comment description, there's the same default as before. But couldn't spot it from a quick read of the code

Never mind, I spotted it. Looks good. Hard to read code on your phone :)

calvinrp (approved):

PR #138, opened 2023-06-21
Symlink proto files to crates

figsoda:
This make sure the crates don't try to access outside their directory, and fixes the interaction with `cargo vendor` when used as a dependency.


peterhuene:
Hi @figsoda.

Thanks for the contribution! We generally try to avoid symlinks as there are folks that like to develop on Windows, and as far as I'm aware, only Windows 11 offers a decent experience with them (i.e. doesn't require administrative rights and enabling "developer mode").

Most of the downstream repositories to this one have been using the warg crates via `git` dependencies in `Cargo.toml`, but not vendoring them.

Is vendoring the warg crates blocking you currently?

We'll need to solve this problem to publish the warg crates to crates.io eventually; perhaps we can consolidate the multiple `build.rs` into a single crate that does the protobuf compilation and exposes the generated Rust types for the other (i.e. `protocol` and `transparency`) crates?

figsoda:
> Is vendoring the warg crates blocking you currently?

I am trying to package cargo-component for nix (https://github.com/NixOS/nixpkgs/pull/239040), which uses warg as a regular git dependency. The issue is that due to the design/limitation of nix, (in nixpkgs) all the dependencies need to be downloaded separately, pretty much equivalent to running `cargo vendor` before building the package, which flattens the workspace into crates and breaks the protobuf compilations.

> perhaps we can consolidate the multiple `build.rs` into a single crate that does the protobuf compilation and exposes the generated Rust types for the other (i.e. `protocol` and `transparency`) crates?

Sounds good to me if that works!

peterhuene:
@Kylebrown9 do you think we should simply have a `warg-protobuf` crate that wraps the output of `protoc` from a single `build.rs` so that we can just share the generated code between transparency and protocol?

That should eliminate the need for symlinks and also make it possible to publish transparency and protocol without having to duplicate the protobuf files into those crates (assuming not using a symlink).

Kylebrown9:
I do like having the protobuf-generated code only in the crates it's needed, but I think a common crate sounds reasonable.

It sounds like we'd need to move the proto files out of the root though, which means we'd lose the "specification stuff is in the root in a common place for all consumers" approach.

figsoda:
It can be at the root if you put the crate also in to proto directory, and to be clear all the non-proto files can be in the top-level of the directory, so it would look something like this
```
proto
├─ google/*/*.proto
├─ warg/*/*.proto
├─ Cargo.toml
├─ README.md
├─ build.rs
└─ lib.rs
```

Kylebrown9:
Not what I would have come up with, but it looks reasonable. 

If I understand correctly the idea is to make a minimal create named e.g. `warg-proto` in the `proto` folder, which would be the only crate that needs the protos and does all the `prost` and `pbjson` codegen. Then we'd add this as a dependency of `warg-transparency` and `warg-protocol`. This would give us the "crate builds only depend on files in their dir structure" invariant.

If that sounds right to you, feel free to take a stab at this.

figsoda:
#139

PR #139, opened 2023-06-24
Move protobuf codegen into its own crate warg-protobuf

figsoda:
closes #138


crates/protocol/src/lib.rs
```diff
@@ -36,11 +36,7 @@ pub trait Validator:
...
 /// Types for converting to and from protobuf
 pub mod protobuf {
-    #![allow(clippy::all)]
-    // Generated by [`prost-build`]
-    include!(concat!(env!("OUT_DIR"), "/warg.protocol.rs"));
-    // Generated by [`pbjson-build`]
-    include!(concat!(env!("OUT_DIR"), "/warg.protocol.serde.rs"));
+    pub use warg_protobuf::protocol::*;
 
     pub fn prost_to_pbjson_timestamp(timestamp: prost_types::Timestamp) -> pbjson_types::Timestamp {
```
lann, line 41: These helpers should presumably migrate to the protobuf crate as well, but that can be done separately.

lann (approved):

PR #144, opened 2023-06-29
Rename validators state

### LINKED ISSUE #94: Refactor some names

lann:
Fixes #94

### DIFF
--- a/crates/client/src/storage.rs
+++ b/crates/client/src/storage.rs
@@ -103,9 +103,9 @@ pub trait ContentStorage: Send + Sync {
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct OperatorInfo {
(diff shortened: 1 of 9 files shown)


calvinrp (approved):

peterhuene (approved):
Looks like this needs a rebase but otherwise GTG.

Issue #145, opened 2023-06-29
`warg key new` fails opening platform secure storage


lann:
@fibonacci1729 

```
error: failed to set signing key with name `default` of registry `http://localhost:8090`: Platform secure storage failure: zbus error: org.freedesktop.Secret.Error.IsLocked: Cannot create an item in a locked collection
```

lann:
After some investigation I believe this is related to running in a headless environment. Since that is likely to be somewhat common for this tool I think we'll need an alternative way to store keys.

rperez95:
Hi @lann 

Any updates on this end?

Today, I installed warg, and when I attempt to set the key for my local repository at 127.0.0.1:8090, I encounter the same error. Subsequently, I installed gnome-keyring using the command:

`sudo apt install gnome-keyring`

Today, I installed warg, and when I attempt to set the key for my local repository at 127.0.0.1:8090, I encounter the same error. Subsequently, I installed gnome-keyring using the command:

```
warg key new 127.0.0.1:8090
error: failed to set signing key with name `default` of registry `https://127.0.0.1:8090/`: Couldn't access platform secure storage: SS error: result not returned from SS API
```
Any idea on how to resolve this?

rperez95:
Hi @lann 

Any updates on this end?

Today, I installed warg, and when I attempt to set the key for my local repository at 127.0.0.1:8090, I encounter the same error. Subsequently, I installed gnome-keyring using the command:

`sudo apt install gnome-keyring`

After installing gnome-keyring, I re-executed the warg command, and now I am encountering the following error:

```
warg key new 127.0.0.1:8090
error: failed to set signing key with name `default` of registry `https://127.0.0.1:8090/`: Couldn't access platform secure storage: SS error: result not returned from SS API
```

rperez95:
Hi @lann 

Any updates on this end?

Today, I installed warg, and when I attempt to set the key for my local repository at 127.0.0.1:8090, I encounter the same error. Subsequently, I installed gnome-keyring using the command:

`sudo apt install gnome-keyring`

After installing gnome-keyring, I re-executed the warg command, and now I am encountering the following error:

```
warg key new 127.0.0.1:8090
error: failed to set signing key with name `default` of registry `https://127.0.0.1:8090/`: Couldn't access platform secure storage: SS error: result not returned from SS API
```

Btw, I'm running warg in a headless environment.

rperez95:
Hi @lann 

Any updates on this end?

[...]

PR #147, opened 2023-06-30
Support loading signing keys from a file

lann:
Not sure this completely solves #145 from a "good UX" perspective, but it does provide a workaround.

Also tightened up handling of sensitive encoded private keys a bit, making sure those get zeroized on drop as well.


crates/crypto/src/signing/private_key.rs
```diff
@@ -17,6 +17,12 @@ pub enum PrivateKeyInner {
 }
 
 impl PrivateKey {
+    /// Read a private key from a file.
+    pub fn from_file(path: impl AsRef<Path>) -> Result<Self, PrivateKeyParseError> {
```
Kylebrown9, line 21: Is there a reason we need to add a `from_file` here and adding IO to the error set?
It feels like we're mixing abstractions and you could just as easily `read_to_string(...).parse()` to keep them separate.

Kylebrown9: My main objection is the `PrivateKeyParseError::IO` addition which isn't actually caused by parsing or reachable when calling `parse`.

lann: I stuck it in here because of `Zeroizing`, but I could probably shuffle things around to avoid it.

fibonacci1729: Just curious bould this just return an `anyhow::Result` so as to not pollute `PrivateKeyParseError`? I find `from_file` conventions to be rather useful for public items like this but fine either way.

lann: Switched `FromStr` -> `TryFrom<String>` which lets it zeroize.

lann: Well you can [see the diff here](https://github.com/bytecodealliance/registry/compare/ed730919cf3f77f95e9e583963916cd8f0cb07c7..8d68811780d8e1b8f639ee8d919c5e4f8b875a62), but github seems to be having trouble with consistency... :grimacing:

crates/crypto/src/signing/private_key.rs
```diff
@@ -58,16 +57,20 @@ impl fmt::Display for PrivateKey {
...
-        let bytes = STANDARD.decode(parts[1])?;
+    fn try_from(key: String) -> Result<Self, PrivateKeyParseError> {
+        let key = Zeroizing::new(key);
+
+        let Some((algo, b64_data)) = key.split_once(':') else {
+            return Err(PrivateKeyParseError::MissingColon)
+        };
+
+        let algo = algo.parse::<SignatureAlgorithm>()?;
+        let bytes = STANDARD.decode(b64_data)?;
```
lann, line 73: Extra colon rejection is handled by base64 decode.

crates/crypto/src/signing/private_key.rs
lann: Had to push an empty diff to convince GH things had actually changed :roll_eyes: .

fibonacci1729 (approved):

lann:
Replaced `FromStr` and `Display` impls with explicit `decode` and `encode` methods, which lets us use zeroize more and makes it less likely to accidentally display the key somewhere.

PR #153, opened 2023-07-06
server: Add /debug/packages endpoint

lann:
This introduces a 'debug' feature to warg-server, which toggles on /debug/-prefixed APIs. The first such api is /debug/packages, which uses a new (optional) DataStore::debug_list_package_ids.

Reorganized some generic API code from `warg_server::api::v1` up to `warg_server::api` to be shared with the new `warg_server::api::debug` module.

### DIFF
--- a/crates/server/src/api/mod.rs
+++ b/crates/server/src/api/mod.rs
@@ -16,6 +16,9 @@ use url::Url;
 
 pub mod v1;
 
+#[cfg(feature = "debug")]
+pub mod debug;
+
 /// Creates the router for the API.
 pub fn create_router(
     content_base_url: Url,
@@ -25,7 +28,10 @@ pub fn create_router(
     content_policy: Option<Arc<dyn ContentPolicy>>,
     record_policy: Option<Arc<dyn RecordPolicy>>,
 ) -> Router {
-    Router::new()
+    let router = Router::new();
+    #[cfg(feature = "debug")]
+    let router = router.nest("/debug", debug::Config::new(core.clone()).into_router());
+    router
         .nest(
             "/v1",
             v1::create_router(
--- a/crates/server/src/datastore/mod.rs
+++ b/crates/server/src/datastore/mod.rs
@@ -259,4 +259,10 @@ pub trait DataStore: Send + Sync {
         log_id: &LogId,
         record: &ProtoEnvelope<package::PackageRecord>,
     ) -> Result<(), DataStoreError>;
+
+    // Returns a list of package names, for debugging only.
+    #[doc(hidden)]
+    async fn debug_list_package_ids(&self) -> anyhow::Result<Vec<PackageId>> {
+        anyhow::bail!("not implemented")
+    }
 }
--- a/crates/server/src/datastore/mod.rs
+++ b/crates/server/src/datastore/mod.rs
@@ -261,6 +261,7 @@ pub trait DataStore: Send + Sync {
     ) -> Result<(), DataStoreError>;
 
     // Returns a list of package names, for debugging only.
+    #[cfg(feature = "debug")]
     #[doc(hidden)]
     async fn debug_list_package_ids(&self) -> anyhow::Result<Vec<PackageId>> {
         anyhow::bail!("not implemented")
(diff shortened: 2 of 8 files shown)


crates/server/src/api/mod.rs
```diff
@@ -63,3 +79,72 @@ pub fn create_router(
                 ),
         )
 }
+
+/// An extractor that wraps the JSON extractor of Axum.
+///
+/// This extractor returns an API error on rejection.
+#[derive(FromRequest)]
+#[from_request(via(axum::Json), rejection(Error))]
+pub struct Json<T>(T);
```
peterhuene, line 88: FYI: the reason these types were in the `v1` mod was that they return `v1` errors. In theory, a `v2` error might have a different format.

crates/server/src/datastore/mod.rs
```diff
@@ -259,4 +259,10 @@ pub trait DataStore: Send + Sync {
         log_id: &LogId,
         record: &ProtoEnvelope<package::PackageRecord>,
     ) -> Result<(), DataStoreError>;
+
+    // Returns a list of package names, for debugging only.
+    #[doc(hidden)]
+    async fn debug_list_package_ids(&self) -> anyhow::Result<Vec<PackageId>> {
+        anyhow::bail!("not implemented")
+    }
 }
```
peterhuene, line 268: Nit: Can we simply call this `get_package_ids` rather than have a `debug_` prefix? It feels like the fact that this is only called from the debug endpoint is leaking into the data store API a little.

Also, could this return `DataStoreError`? If this ever ends up being called from a "non-debug" endpoint, it would be nice for callers to be able to differentiate between an error like the postgres database is down (and return a 500 with no context) vs. something else.

crates/server/src/datastore/postgres/mod.rs
```diff
@@ -757,4 +757,17 @@ impl DataStore for PostgresDataStore {
         package::PackageRecord::verify(key, record.content_bytes(), record.signature())
             .map_err(|_| DataStoreError::SignatureVerificationFailed)
     }
+
+    async fn debug_list_package_ids(&self) -> anyhow::Result<Vec<PackageId>> {
+        let mut conn = self.0.get().await?;
+        let names = schema::logs::table
+            .select(schema::logs::name)
+            .load::<Option<String>>(&mut conn)
```
peterhuene, line 765: We can use `Text<PackageId>` here I think.

peterhuene (approved):
Generally looks great, just a few comments!

crates/server/src/datastore/mod.rs
lann: I'd be happy to take a different approach, but the issue is that this method just operationally doesn't make sense for production; even if there is some production use case for getting all package IDs, you'd probably want this to return a `Stream`, which is unnecessarily complicated to implement for a debug interface.

crates/server/src/datastore/postgres/mod.rs
lann: Ah nice, will do.

crates/server/src/datastore/mod.rs
peterhuene: That's a good point re: streaming.

Should it be `cfg(feature = "debug")` to ensure it isn't normally present on the trait (and impls) given its inefficiencies, then?

crates/server/src/api/mod.rs
lann: Makes sense. I'll revert and have the debug API depend on v1 explicitly; it just requires making some private fields `pub(crate)`.

crates/server/src/datastore/mod.rs
lann: Ah yes, that does make sense.

crates/server/src/datastore/postgres/mod.rs
lann: This ended up being a hassle because `PackageId`'s `FromStr:Err` doesn't match what `ParsedText` wants.

peterhuene: Ah yeah, that makes sense. Can leave as-is then.

peterhuene (approved):

PR #157, opened 2023-07-10
Support authorized keys policy config file

lann:
- Add `warg-server --authorized-keys-file`
- Add `warg key id` subcommand to generate authorized key values


src/commands/key.rs
```diff
@@ -60,35 +98,49 @@ impl KeyNewCommand {
...
         let key = SigningKey::random(&mut OsRng).into();
-        set_signing_key(&self.host, &self.key_name, &key)?;
+        self.keyring_entry.set_entry(&key)?;
 
         Ok(())
     }
 }
 
+/// Shows the ID of the signing key for a registry in the local keyring.
+
```
peterhuene, line 121:
```suggestion
```

src/commands/key.rs
```diff
@@ -60,35 +98,49 @@ impl KeyNewCommand {
...
+#[derive(Args)]
+pub struct KeyIdCommand {
+    
+
```
peterhuene, line 125:
```suggestion
```

PR #158, opened 2023-07-10
Fix warg update

lann:


crates/client/src/storage/fs.rs
```diff
@@ -95,11 +96,12 @@ impl RegistryStorage for FileSystemRegistryStorage {
     async fn load_packages(&self) -> Result<Vec<PackageInfo>> {
         let mut packages = Vec::new();
 
-        for entry in WalkDir::new(&self.base_dir) {
+        let packages_dir = self.base_dir.join(PACKAGE_LOGS_DIR);
+        for entry in WalkDir::new(&packages_dir) {
```
lann, line 100: The minimal fix for `warg update` would have been

```suggestion
        for entry in WalkDir::new(&packages_dir.join("sha256")) {
```

but given how packages are iterated over this seemed more robust.

crates/client/src/storage/fs.rs
```diff
@@ -70,10 +71,10 @@ impl FileSystemRegistryStorage {
     }
 
     fn package_path(&self, id: &PackageId) -> PathBuf {
-        self.base_dir.join(
+        self.base_dir.join(PACKAGE_LOGS_DIR).join(
             LogId::package_log::<Sha256>(id)
                 .to_string()
-                .replace(':', "/"),
+                .replace(':', "-"),
```
peterhuene, line 77: As `WalkDir` is recursive, can we keep the algo directory?

peterhuene (approved):

peterhuene:
Also probably need a client test to cover the update command?

crates/client/src/storage/fs.rs
lann: Sure.

lann:
Added an assertion to the client megatest & confirmed it would have caught the regression.

peterhuene (approved):

PR #161, opened 2023-07-12
corrected openapi.yaml to match the current implementation

calvinrp:
See: https://github.com/bytecodealliance/registry/blob/main/crates/api/src/v1/fetch.rs#L18

### DIFF
--- a/crates/server/openapi.yaml
+++ b/crates/server/openapi.yaml
@@ -496,7 +496,7 @@ components:
       description: A request to fetch logs from the registry.
       additionalProperties: false
       required:
-        - checkpoint
+        - root
       properties:
         root:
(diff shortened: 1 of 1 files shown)


crates/server/openapi.yaml
```diff
@@ -498,9 +498,9 @@ components:
       required:
         - checkpoint
```
peterhuene, line 499:
```suggestion
        - root
```

peterhuene (approved):
Just need the update to the required properties and 👍

lann (approved):

calvinrp:
I don't feel strongly about this. This PR would be an "either or", comparing against: https://github.com/bytecodealliance/registry/pull/165

PR #165, opened 2023-07-13
rename `root` field to `checkpoint_id` for the API call to fetch logs

calvinrp:
`checkpoint_id` is probably more clear and consistent than `root` or `checkpoint`. Breaking change while they are still easy.

I don't feel strongly about this. This PR would be an "either or", comparing against: https://github.com/bytecodealliance/registry/pull/161


lann (approved):
Less terms is more better.

Kylebrown9:
Seems reasonable assuming `checkpoint_id` is the hash of a checkpoint.

It might be time to do some more glossary work.

PR #166, opened 2023-07-13
Add 'package_keys' to AuthorizedKeyPolicy

lann:
This allows signing keys to be granted write access to a log based on the full package name.


peterhuene (approved):
Looks good!

PR #146, opened 2023-06-29
Abstract content storage via trait and introduce OCI content storage

devigned:
This PR introduces the concept of a "ContentStore", which is an abstraction for storing package content. There are currently 2 implementations for a content store, local and OCI. However, it would be relatively simple to add another for a blob store like S3, Azure Blobs, R2, etc.

In this PR, the default file storage and serving directory is replaced with URIs that are served via the LocalContentStore. It should provide the same functionality as the current local file store / serving.

There is still some work to be done with the current OCI implementation to fill the image config with details more than just empty JSON. Furthermore, since the PR is using the image config media type, the artifactType must be specified, and I don't think that is yet available in oci-distribution lib. I will follow up on both of these details.

I'm opening this PR to start a discussion of implementation that will help to fill in the blanks for #87. Also, it's my first significant contribution to this repo, so I'd really appreciate guidance. Thank you!!

### Start a local registry
```
docker run -d -p 5000:5000 --restart=always --name registry registry:2
```

### Start warg-server 
Tell warg-server to use the OCI v1.1 content store. It will default to connecting to a registry insecurely at localhost:5000
```
david at hulkbuster in ~/code/bca/registry on oci-cs [$]
$ WARG_CONTENT_STORE=ociv1-1 cargo run -p warg-server -- --content-dir content
```

### Publish a demo package
```
david at hulkbuster in ~/code/bca/registry on oci-cs [$]
$ warg publish init default:hello
published initialization of package `default:hello`

david at hulkbuster in ~/code/bca/registry on oci-cs [$]
$ warg publish release -i default:hello -v 0.1.0 demo/simple-grep-1.0.0.wasm
published version 0.1.0 of package `default:hello`
```

### Verify the content is stored in the local registry
```
david at hulkbuster in ~/code/bca/registry on oci-cs [$]
$ regctl repo ls localhost:5000
WARN[0000] Changing credential host for registry         host=index.docker.io new="[https://index.docker.io/v1//refresh-token"](https://index.docker.io/v1//refresh-token%22) orig="[https://index.docker.io/v1//access-token"](https://index.docker.io/v1//access-token%22)
default/hello

david at hulkbuster in ~/code/bca/registry on oci-cs [$]
$ regctl manifest get localhost:5000/default/hello:0.1.0
[...]

PR #137, opened 2023-06-15
use missing url source for content upload

macovedj:

PR #167, opened 2023-07-14
Update missing content API

lann:
This implements the proposal here: https://github.com/bytecodealliance/registry/discussions/160 It adapts _some_ of #137. Notably it does _not_ add a "patch record" operation as I think that conflicts with #146.

I also fell down a bit of a rabbit hole to allow upload URLs to be specified relative to the registry URL which led to a refactor that [adds a `RegistryUrl` type](https://github.com/bytecodealliance/registry/pull/167/commits/30d900f0d9a39c591c9d0864efd53ace31e6ccd7). Note that this will (probably) break existing client config for `localhost` environments as it makes port numbers significant for on-disk paths and the keyring.

### DIFF
--- a/crates/api/src/v1/paths.rs
+++ b/crates/api/src/v1/paths.rs
@@ -1,6 +1,5 @@
 //! The paths of the Warg REST API.
 
-use warg_crypto::hash::AnyHash;
 use warg_protocol::registry::{LogId, RecordId};
 
 /// The path of the "fetch logs" API.
@@ -21,11 +20,6 @@ pub fn publish_package_record(log_id: &LogId) -> String {
 /// The path for a package record.
 pub fn package_record(log_id: &LogId, record_id: &RecordId) -> String {
     format!("v1/package/{log_id}/record/{record_id}")
-}
-
-/// The path for a package record's content.
-pub fn package_record_content(log_id: &LogId, record_id: &RecordId, digest: &AnyHash) -> String {
-    format!("v1/package/{log_id}/record/{record_id}/content/{digest}")
 }
 
 /// The path for proving checkpoint consistency.
--- a/crates/client/src/registry_url.rs
+++ b/crates/client/src/registry_url.rs
@@ -1,0 +1,194 @@
+use anyhow::{anyhow, bail, Context, Result};
+use reqwest::IntoUrl;
+use url::{Host, Url};
+
+/// The base URL of a registry server.
+// Note: The inner Url always has a scheme and host.
+#[derive(Clone)]
+pub struct RegistryUrl(Url);
+
+impl RegistryUrl {
+    /// Parses and validates the given URL into a [`RegistryUrl`].
+    pub fn new(url: impl IntoUrl) -> Result<Self> {
+        // Default to a HTTPS scheme if none is provided
+        let mut url: Url = if !url.as_str().contains("://") {
+            Url::parse(&format!("https://{url}", url = url.as_str()))
+                .context("failed to parse registry server URL")?
+        } else {
+            url.into_url()
+                .context("failed to parse registry server URL")?
+        };
+
+        match url.scheme() {
+            "https" => {}
+            "http" => {
+                // Only allow HTTP connections to loopback
+                match url
+                    .host()
+                    .ok_or_else(|| anyhow!("expected a host for URL `{url}`"))?
+                {
+                    Host::Domain(d) => {
+                        if d != "localhost" {
+                            bail!("an unsecured connection is not permitted to `{d}`");
+                        }
+                    }
+                    Host::Ipv4(ip) => {
+                        if !ip.is_loopback() {
+                            bail!("an unsecured connection is not permitted to address `{ip}`");
+                        }
+                    }
+                    Host::Ipv6(ip) => {
+                        if !ip.is_loopback() {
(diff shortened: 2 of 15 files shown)


crates/server/openapi.yaml
```diff
@@ -700,12 +700,9 @@ components:
           enum: [sourcing]
           example: sourcing
         missingContent:
-          type: array
-          description: The array of content digests that are missing for the package record.
+          "$ref": "#/components/schemas/MissingContentMap"
+          description: The missing content for the package record.
           minItems: 1
```
peterhuene, line 705:
```suggestion
```

peterhuene: Also, we can probably remove `missingContent` from the required properties; a `null` value would indicate no missing content.

crates/server/openapi.yaml
```diff
@@ -825,6 +822,44 @@ components:
...
+    MissingContentMap:
+      type: object
+      description: The map of content digest to missing content info.
+      patternProperties:
+        "^[a-z0-9-]+:[a-f0-9]+$":
+          "$ref": "#/components/schemas/MissingContent"
+      example:
+        ? "sha256:7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded9773"
+        : upload:
+            - type: http-post
```
peterhuene, line 834: We should keep to a JSON naming convention for enum values:

```suggestion
            - type: httpPost
```

crates/server/openapi.yaml
```diff
@@ -825,6 +822,44 @@ components:
...
+        upload:
+          description: Upload endpoint(s) for the missing content.
+          type: array 
+          items:
+            oneOf:
+              - "$ref": "#/components/schemas/HttpPostUpload"
+            discriminator:
+              propertyName: type
+              mapping:
+                http-post: "#/components/schemas/HttpPostUpload"
```
peterhuene, line 848: See comment above:

```suggestion
                httpPost: "#/components/schemas/HttpPostUpload"
```

crates/server/openapi.yaml
```diff
@@ -825,6 +822,44 @@ components:
...
+    HttpPostUpload:
+      type: object
+      description: A HTTP POST upload endpoint.
+      properties:
+        type:
+          type: string
+          description: The type of upload endpoint.
+          enum: [http-post]
+          example: http-post
```
peterhuene, line 857:
```suggestion
          enum: [httpPost]
          example: httpPost
```

peterhuene (commented):
My feedback on just `openapi.yaml`. Haven't looked over the rest of the changes yet.

crates/server/openapi.yaml
lann: Is it valid to be in `state: 'sourcing'` with no `missingContent`?

peterhuene: It shouldn't be; right now the server will transition directly to the `processing` state if nothing is missing.

peterhuene: Oh, yeah, then I guess keeping `minItems` is fine, so disregard that comment.

peterhuene: That said, I'm not entirely sure if `minItems` is valid on an object (the OpenAPI documentation mentions it only for array types).

lann: It's `minProperties`, apparently.

crates/client/src/registry_url.rs
```diff
@@ -0,0 +1,194 @@
...
+                    }
+                    Host::Ipv6(ip) => {
+                        if !ip.is_loopback() {
+                            bail!("an unsecured connection is not permitted to address `{ip}`");
+                        }
+                    }
+                }
+            }
+            _ => bail!("expected a HTTPS scheme for URL `{url}`"),
+        }
```
lann, line 48: Above here is moved straight from `Client::validate_url`

crates/client/src/registry_url.rs
```diff
@@ -0,0 +1,194 @@
...
+            url.set_path(&(url.path().to_string() + "/"));
+        }
+
+        Ok(Self(url))
+    }
+
+    /// Returns a mostly-human-readable string that identifies the registry and
+    /// contains only the characters `[0-9a-zA-Z-._]`. This string is
+    /// appropriate to use with external systems that can't accept arbitrary
+    /// URLs such as file system paths.
```
lann, line 61: See tests for examples.

crates/api/src/v1/paths.rs
```diff
@@ -23,11 +22,6 @@ pub fn package_record(log_id: &LogId, record_id: &RecordId) -> String {
     format!("v1/package/{log_id}/record/{record_id}")
 }
 
-/// The path for a package record's content.
-pub fn package_record_content(log_id: &LogId, record_id: &RecordId, digest: &AnyHash) -> String {
-    format!("v1/package/{log_id}/record/{record_id}/content/{digest}")
-}
```
lann, line 29: Moved into the server crate as it is no longer part of the API spec.

crates/client/src/registry_url.rs
peterhuene: Thanks for catching this issue in the client!

peterhuene (approved):
Looks good! 👍

PR #169, opened 2023-07-19
CoreService refactor

lann:
This flattens all of the (other) server services into `CoreService`. I've mostly kept the external interfaces the same for now, with some notable exceptions:

- `CoreService` now contains an `Arc<Inner>`, so no longer needs a wrapping `Arc`
- Since the `log` and `map` services went away, so too have `CoreService::log_data()` and `::map_data()`. The related proof methods have been flattened into `CoreService` as e.g. `::log_consistency_proof`

The `core.rs` diff is completely useless; see just the new version here: https://github.com/bytecodealliance/registry/blob/core-svc-refactor/crates/server/src/services/core.rs

The new implementation reuses a lot of the existing services' code but does make some functional changes:

- Rather than attempting to write a final checkpoint on shutdown, a new checkpoint may be written just after initialization, which will also help recover from non-graceful shutdowns.
- Locks are all handled internally now, rather than some being held by callers.
- The `CancellationToken` has gone away; the service can now be stopped by dropping all references to it. This also meant the `StopHandle` could be reduced to a normal tokio `JoinHandle` which can be `await`ed to give the record processing loop a chance to finish any queued updates.


crates/crypto/src/hash/mod.rs
```diff
@@ -80,7 +80,7 @@ where
     Hash::of((0b1, lhs, rhs))
 }
 
-pub trait SupportedDigest: Digest + private::Sealed + Sized + 'static {
+pub trait SupportedDigest: Digest + private::Sealed + Default + Sized + 'static {
```
lann, line 83: `Sha256` already implements `Default`; this just makes `#[derive(Default)]` work more places.

calvinrp (approved):
Much simpler. Really good work refactoring tricky code. Looks good to me, but I'd love @peterhuene review as well.

crates/server/src/services/core.rs
```diff
@@ -1,297 +1,395 @@
...
+            // Push the entry itself
+            state.push_entry(leaf.clone());
+
+            // Update per-checkpoint state and track any entries that aren't
+            // yet part of a checkpoint
+            if let Some(checkpoint) = checkpoint {
+                if last_checkpoint != checkpoint {
+                    state.checkpoint(); // for the side-effect of updating map_index
+                    last_checkpoint = checkpoint;
                 }
```
peterhuene, line 206: I think this contains a bug copied over from my initial implementation of server restart that I haven't gotten around to fixing.

Let's say the initial stream looks like (in record insert order):

| Leaf | Checkpoint|
|------|------|
| A  |  C1 |
| B | C2 |
| C | C3 |
| D | C3 |
| E | C3 |
| F | C4 | 

I think this code would correctly create a checkpoint `C1` containing `A` (`C1 != <empty-hash>`) and a checkpoint `C2` containing `B` (`C2 != C1`), but would not correctly checkpoint `C3` because we're checkpointing at the _transition_ from leaf `B` to `C` (`C3 != C2`) in the stream, which results in only including leaf `C` in the checkpoint.

The next checkpoint would then incorrectly include leafs `D`, `E`, and `F` (`C4 != C3`).

Does that make sense or am I completely off-base?

peterhuene (commented):
Other than what I believe to be that existing bug, these changes look really good and make this much easier to wrap one's brain around.

crates/server/src/services/core.rs
lann: Ah yeah I was a little confused about this but wanted to stick with existing behavior. I'm currently working on a followup refactor that avoids this problem entirely, but I can retrofit this PR if you'd prefer.

peterhuene: I'll approve and leave it up to you whether you want to include it now or wait to the refactor.

peterhuene (approved):

lann:
OK, since this is such a big diff I'm going to get it out of the way.

PR #171, opened 2023-07-21
Replace record checkpoint field with registry log index

lann:
Prior to this commit, every published record is associated with the first checkpoint created after the record was added to the registry log. While this does aid clients in verifying the inclusion of a record, it is more information than necessary and forces the registry to hold on to all checkpoints indefinitely. It also requires additional bookkeeping just to track this partially-unnecessary information.

This removes that per-record checkpoint and instead associates each record with its index in the registry log. This is still enough information to aid inclusion proofs and reduces the overall complexity of the server.

### DIFF
--- a/crates/server/src/datastore/mod.rs
+++ b/crates/server/src/datastore/mod.rs
@@ -63,17 +63,6 @@ pub enum DataStoreError {
     Diesel(#[from] diesel::result::Error),
 }
 
-/// Represents a leaf used to initialize the core service.
-pub struct InitialLeaf {
-    /// The log leaf.
-    pub leaf: LogLeaf,
-    /// The checkpoint of the leaf.
-    ///
-    /// A value of `None` indicates the leaf needs to be included in the
-    /// next checkpoint.
-    pub checkpoint: Option<AnyHash>,
-}
-
 /// Represents the status of a record.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum RecordStatus {
@@ -98,24 +87,31 @@ where
     pub status: RecordStatus,
     /// The envelope containing the record contents.
     pub envelope: ProtoEnvelope<T>,
-    /// The checkpoint of the record.
+    /// The index of the record in the registry log.
     ///
     /// This is `None` if the record is not published.
-    pub checkpoint: Option<SerdeEnvelope<MapCheckpoint>>,
+    pub registry_log_index: Option<u64>,
 }
 
 /// Implemented by data stores.
 #[axum::async_trait]
 pub trait DataStore: Send + Sync {
-    /// Gets a stream of initial leaves in the store.
+    /// Gets a stream of all checkpoints.
     ///
     /// This is an expensive operation and should only be performed on startup.
-    async fn get_initial_leaves(
+    async fn get_all_checkpoints(
         &self,
     ) -> Result<
-        Pin<Box<dyn Stream<Item = Result<InitialLeaf, DataStoreError>> + Send>>,
+        Pin<Box<dyn Stream<Item = Result<MapCheckpoint, DataStoreError>> + Send>>,
         DataStoreError,
     >;
+
+    /// Gets a stream of all validated records.
+    ///
+    /// This is an expensive operation and should only be performed on startup.
+    async fn get_all_validated_records(
+        &self,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<LogLeaf, DataStoreError>> + Send>>, DataStoreError>;
 
     /// Stores the given operator record.
     async fn store_operator_record(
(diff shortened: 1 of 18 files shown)


crates/server/src/datastore/postgres/migrations/2023-04-11-013700_initial_db/up.sql
```diff
@@ -33,7 +33,7 @@ CREATE TABLE records (
   id SERIAL PRIMARY KEY,
   log_id INTEGER NOT NULL REFERENCES logs(id),
   record_id TEXT NOT NULL UNIQUE,
-  checkpoint_id INTEGER REFERENCES checkpoints(id),
+  registry_log_index BIGINT UNIQUE,
```
lann, line 36: This could have been a separate migration, but it would have been a significant effort to actually migrate the data. Since no one yet has production data it seems reasonable to rebuild databases for now.

crates/server/src/datastore/mod.rs
```diff
@@ -135,15 +131,16 @@ pub trait DataStore: Send + Sync {
...
     ) -> Result<(), DataStoreError>;
 
-    /// Validates the given operator record.
+    /// Commits the given operator record.
     ///
     /// The record must be in a pending state.
     ///
     /// If validation succeeds, the record will be considered part of the log.
-    async fn validate_operator_record(
+    async fn commit_operator_record(
```
lann, line 139: This is somewhat of an unrelated change but I think this name better captures that the method is expected to _update_ the store, not just validate.

crates/server/src/datastore/postgres/mod.rs
```diff
@@ -695,10 +696,12 @@ impl DataStore for PostgresDataStore {
             .first::<Checkpoint>(&mut conn)
             .await?;
 
+        let log_length = checkpoint.log_length.try_into().unwrap();
```
lann, line 699: Hopefully this code will be long gone by the time anyone overflows a u32-worth of records but I'd still rather blow up if it happens.

crates/server/src/services/core.rs
```diff
@@ -176,41 +176,37 @@ struct Inner<Digest: SupportedDigest> {
...
-                pending_entries.clear();
-            } else {
-                pending_entries.push(leaf);
+        let state = self.state.get_mut();
+        while let Some(entry) = published.next().await {
+            state.push_entry(entry?);
+            if let Some(stored_checkpoint) = checkpoints_by_len.get(&state.log.length()) {
+                // Validate stored checkpoint (and update internal state as a side-effect)
+                let computed_checkpoint = state.checkpoint();
+                assert!(stored_checkpoint == &computed_checkpoint);
```
lann, line 205: @peterhuene I believe this avoids the checkpoint issue we discussed, and additionally validates the stored checkpoints.

crates/server/src/datastore/postgres/migrations/2023-04-11-013700_initial_db/up.sql
peterhuene: Pinging @calvinrp here just so he's aware that we'll need to reset the database if this change is deployed to his dogfood server.

crates/server/src/lib.rs
```diff
@@ -140,22 +144,15 @@ impl Config {
...
-            listener: None,
-        }
+        Self { config }
     }
 
-    /// Binds the server to the configured address.
-    ///
-    /// Returns the address the server bound to.
-    pub fn bind(&mut self) -> Result<SocketAddr> {
+    pub async fn initialize(self) -> Result<InitializedServer> {
```
peterhuene, line 155: Nit: missing doc comment.

crates/server/src/lib.rs
```diff
@@ -217,25 +197,53 @@ impl Server {
...
-        tracing::info!("listening on {addr}", addr = self.config.addr.unwrap());
+        Ok(InitializedServer {
+            listener,
+            router,
+            core_handle,
+            shutdown: self.config.shutdown,
+        })
+    }
+
+    pub async fn run(self) -> Result<()> {
```
peterhuene, line 222: Nit: missing doc comment.

crates/server/src/lib.rs
```diff
@@ -217,25 +197,53 @@ impl Server {
...
+        self.initialize().await?.serve().await
+    }
+}
 
-        if let Some(shutdown) = self.config.shutdown {
+pub struct InitializedServer {
```
peterhuene, line 227: Nit: missing doc comment (aside: perhaps we should slap a `#![deny(missing_docs)]` on this crate since it's intended to be used as a library?).

crates/server/src/lib.rs
```diff
@@ -217,25 +197,53 @@ impl Server {
...
+    listener: TcpListener,
+    router: Router,
+    core_handle: JoinHandle<()>,
+    shutdown: Option<ShutdownFut>,
+}
+
+impl InitializedServer {
+    pub fn local_addr(&self) -> std::io::Result<SocketAddr> {
```
peterhuene, line 235: Nit: missing doc comment.

crates/server/src/lib.rs
```diff
@@ -217,25 +197,53 @@ impl Server {
...
+        self.listener.local_addr()
+    }
+
+    pub async fn serve(self) -> Result<()> {
```
peterhuene, line 239: Nit: missing doc comment.

peterhuene (approved):
Really great work! This both simplified things a bit and eliminated that restart checkpoint bug I mentioned earlier.

Just a few nit comments.

PR #172, opened 2023-07-21
Base URL for Warg API change from `/v1/` to `/warg/v1`

calvinrp:
It will likely be common to configure a Warg registry with just a domain name (`warg.example.com` without an explicit path prefix) and that domain might have other non-Warg APIs for UIs or other services. It will be helpful to namespace the Warg APIs a bit with an additional prefix `/warg` in addition to the API version `/v1` prefix.


lann:
> that domain might have other non-Warg APIs for UIs or other services

I would caution against doing this. It is very easy to make subtle mistakes around same-origin security restrictions in services if you have services share domains.

calvinrp:
> > that domain might have other non-Warg APIs for UIs or other services
> 
> I would caution against doing this. It is very easy to make subtle mistakes around same-origin security restrictions in services if you have services share domains.

Understand your point but I'm thinking of a registry UI that would both use Warg API endpoints and other API endpoints that aren't apart of the Warg protocol. Happy to discuss further. It is more of segmenting Warg API vs other registry-related APIs.

peterhuene:
I feel like we already have the protocol version prefix, so "extension" endpoints, if they must be colocated with the warg API, can exist under a different subpath without conflicting with the warg API, no?

lann:
As a general practice I would try to avoid running services that live in separate code bases on the same domain. The web security model works best if separate services are served from separate origins, _especially_ if user-generated content is in play, which is why you see e.g. Github raw files served from `raw.githubusercontent.com`.

lann:
Actually, we already configure registries by full URL, allowing a particular registry to add a path prefix if desired. We'll need to make sure that path actually works in the client code, but it seems like it would address this and similar needs.

calvinrp:
> Actually, we already configure registries by full URL, allowing a particular registry to add a path prefix if desired. We'll need to make sure that path actually works in the client code, but it seems like it would address this and similar needs.

Yes, the `warg config --registry example.com/somepath` should work today. This PR isn't a blocker for me. But I think it merits a discussion at our next Registry meeting on Wed.

calvinrp:
Abandoning this PR. Can revisit.

PR #174, opened 2023-07-25
server: Add /debug/package/<id> endpoint

lann:
This dumps a package's log and derived (validated) state as a debugging tool.


crates/server/src/api/debug/mod.rs
```diff
@@ -18,18 +35,144 @@ impl Config {
     pub fn into_router(self) -> Router {
         Router::new()
             .route("/packages", get(list_package_names))
+            .route("/package/:name", get(get_package_info))
```
peterhuene, line 38: Nit: while name is pretty synonymous with package id, perhaps this just for clarity?

```suggestion
            .route("/package/:package_id", get(get_package_info))
```

peterhuene (approved):
Just the one minor nit.

crates/server/src/api/debug/mod.rs
lann: Oops, I even changed it in the function signature...

PR #184, opened 2023-08-21
prove consistency endpoint using log_lengths instead of log_roots

calvinrp:


lann (approved):
I do think we should use e.g. `log_length` and `leaf_indexes` in `ConsistencyRequest` for...consistency...

PR #186, opened 2023-08-24
Support delegation in AuthorizedKeyPolicy

### LINKED ISSUE #185: Support publishing from "grant"ed publishers
The current policy is a allowlist for _all_ publishes. We should loosen this to allow grantees to publish for componentize the world.

lann:
This adds a `delegation` flag to AuthorizedKeyPolicy which is configurable per-namespace/package. When enabled, the authorized key enforcement for that scope only applies to init messages.

To support the flag, the authorized keys toml format has changed to e.g.:

```toml
[namespace.foo]
keys = ["..."]
delegation = true
```

Fixes #185

Also a couple of formatting changes from rustfmt 1.6 since we enforce latest stable.

### DIFF
--- a/crates/protocol/src/package/state.rs
+++ b/crates/protocol/src/package/state.rs
@@ -386,6 +386,12 @@ impl LogState {
             return Err(ValidationError::InitialEntryAfterBeginning);
         }
 
+        if signer_key_id != &init_key.fingerprint() {
+            return Err(ValidationError::KeyIDNotRecognized {
+                key_id: signer_key_id.clone(),
+            });
+        }
+
         assert!(self.permissions.is_empty());
         assert!(self.releases.is_empty());
         assert!(self.keys.is_empty());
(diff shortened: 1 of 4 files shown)


crates/protocol/src/package/state.rs
```diff
@@ -386,6 +386,12 @@ impl LogState {
             return Err(ValidationError::InitialEntryAfterBeginning);
         }
 
+        if signer_key_id != &init_key.fingerprint() {
+            return Err(ValidationError::KeyIDNotRecognized {
+                key_id: signer_key_id.clone(),
+            });
+        }
```
lann, line 393: This requires that the public key in the init entry (which is effectively a "grant") matches the signing key. Not strictly necessary for this PR but seems like a good conservative check for now.

peterhuene (approved):
Thanks for fixing this!

PR #187, opened 2023-08-25
added registry index for fetched records and updated proof endpoints to use the log length and registry index

calvinrp:
The registry index for a published record is going to be different than the `VecLog`'s `Node(usize)` value that is required for the log inclusion proof. For now, keeping the existing pattern of using a `HashMap` (could be a `Vec` if added back in the same order) to lookup the value in memory.

To make it easier to adjust the size of the integer for the `log_length` and `registry_index`, I added a type alias `RegistryIndex`.

As always, suggestions appreciated. This was more of a change than I expected.

### DIFF
--- a/crates/api/src/v1/fetch.rs
+++ b/crates/api/src/v1/fetch.rs
@@ -144,9 +144,11 @@ impl<'de> Deserialize<'de> for FetchError {
         D: serde::Deserializer<'de>,
     {
         match RawError::<String>::deserialize(deserializer)? {
-            RawError::CheckpointNotFound { status: _, ty: _, id } => {
-                Ok(Self::CheckpointNotFound(id))
-            },
+            RawError::CheckpointNotFound {
+                status: _,
+                ty: _,
+                id,
+            } => Ok(Self::CheckpointNotFound(id)),
             RawError::NotFound { status: _, ty, id } => match ty {
                 EntityType::Log => Ok(Self::LogNotFound(
                     id.parse::<AnyHash>()
@@ -166,9 +168,9 @@ impl<'de> Deserialize<'de> for FetchError {
                         .into(),
                 )),
                 _ => Err(serde::de::Error::invalid_value(
-                                Unexpected::Str(&id),
-                                &"a valid log length",
-                            )),
+                    Unexpected::Str(&id),
+                    &"a valid log length",
+                )),
             },
             RawError::Message { status, message } => Ok(Self::Message {
                 status,
--- a/crates/api/src/v1/fetch.rs
+++ b/crates/api/src/v1/fetch.rs
@@ -6,7 +6,7 @@ use std::{borrow::Cow, collections::HashMap};
 use thiserror::Error;
 use warg_crypto::hash::AnyHash;
 use warg_protocol::{
-    registry::{LogId, RecordId},
+    registry::{LogId, RecordId, RegistryLen},
     PublishedProtoEnvelopeBody,
 };
 
@@ -14,8 +14,8 @@ use warg_protocol::{
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct FetchLogsRequest<'a> {
-    /// The root checkpoint ID hash of the registry.
-    pub checkpoint_id: Cow<'a, AnyHash>,
+    /// The checkpoint log length.
+    pub log_length: RegistryLen,
     /// The limit for the number of operator and package records to fetch.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub limit: Option<u16>,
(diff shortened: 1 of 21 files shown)


calvinrp:
I need to investigate why it is failing on the inclusion proof on the Postgres test. Trying to reproduce.

crates/api/src/v1/proof.rs
```diff
@@ -31,11 +31,11 @@ pub struct ConsistencyResponse {
...
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct InclusionRequest<'a> {
-    /// The checkpoint to check for inclusion.
-    pub checkpoint: Cow<'a, Checkpoint>,
-    /// The log leafs to check for inclusion.
-    pub leafs: Cow<'a, [LogLeaf]>,
+pub struct InclusionRequest {
+    /// The log length to check for inclusion.
+    pub log_length: RegistryIndex,
```
lann, line 36: It would be _nice_ to give this a separate type name e.g. `RegistryLen`, but not critical.

crates/server/src/api/v1/proof.rs
```diff
@@ -27,20 +28,13 @@ impl Config {
...
-
 impl From<CoreServiceError> for ProofApiError {
     fn from(value: CoreServiceError) -> Self {
         Self(match value {
-            CoreServiceError::RootNotFound(root) => ProofError::RootNotFound(root),
-            CoreServiceError::LeafNotFound(leaf) => ProofError::LeafNotFound(leaf),
+            CoreServiceError::CheckpointNotFound(log_length) => {
+                ProofError::CheckpointNotFound(log_length)
+            }
+            //CoreServiceError::LeafNotFound(leaf) => ProofError::LeafNotFound(leaf),
```
lann, line 37: Intentional?

lann (commented):
Looks pretty good to me, pending CI failure.

crates/api/src/v1/proof.rs
calvinrp: Do you think better to have two type aliases? One for the log length and the other for indexes into the log? Or one with a better name?

lann: Two, since you can't generally use one where you need the other.

calvinrp:
Should be good to go now. 

The Postgres test failure was tricky to track down, since I thought it was failing on the inclusion of the log. It was the map inclusion proof that was failing intermittently. When constructing the proof bundle, the leafs need to be added in the exact same order as the client provided the log leafs. It was easier to fix the ordering in Rust than mess with the `diesel` Rust Postgres query.

calvinrp:
Also, updated the `/fetch/logs` endpoint to use `logLength` instead of `checkpointId`. Hopefully, not a controversial change. It should make things easier.

The updated rendered [OpenAPI spec](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/bytecodealliance/registry/163369c0b75d7d627a3f891a8789a5002ca7283a/crates/server/openapi.yaml).

lann (commented):
Could you give this a pass through `cargo clippy`? If you use vscode you can make it the default linter with `"rust-analyzer.check.command": "clippy"`

crates/server/src/datastore/mod.rs
```diff
@@ -111,7 +116,16 @@ pub trait DataStore: Send + Sync {
     /// This is an expensive operation and should only be performed on startup.
     async fn get_all_validated_records(
         &self,
-    ) -> Result<Pin<Box<dyn Stream<Item = Result<LogLeaf, DataStoreError>> + Send>>, DataStoreError>;
+    ) -> Result<
+        Pin<Box<dyn Stream<Item = Result<(RegistryIndex, LogLeaf), DataStoreError>> + Send>>,
```
lann, line 120: Isn't the `RegistryIndex` guaranteed to always be `0..N`?

crates/api/src/v1/fetch.rs
```diff
@@ -138,15 +144,12 @@ impl<'de> Deserialize<'de> for FetchError {
         D: serde::Deserializer<'de>,
     {
         match RawError::<String>::deserialize(deserializer)? {
+            RawError::CheckpointNotFound {
+                status: _,
+                ty: _,
+                id,
```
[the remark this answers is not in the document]
calvinrp: Thanks for catching that particular one. Cleaner. Fixing. Clippy didn't catch that one. Clippy caught others though.

lann (approved):
Not sure I understand the new `RegistryLen` param threaded through `CoreService` but otherwise lgtm.

crates/protocol/src/registry.rs
```diff
@@ -9,11 +9,17 @@ use warg_crypto::prefix::VisitPrefixEncode;
 use warg_crypto::{prefix, ByteVisitor, Signable, VisitBytes};
 use wasmparser::names::KebabStr;
 
+/// Type alias for registry log index
+pub type RegistryIndex = usize;
+
+/// Type alias for registry log length
+pub type RegistryLen = RegistryIndex;
```
lann, line 16: :+1: thanks

crates/server/src/datastore/mod.rs
lann: I think so? You can always get the index later with e.g. https://docs.rs/futures/0.3.28/futures/stream/trait.StreamExt.html#method.enumerate if you need it.

lann: ...though if I'm reading the code correctly I think `State` could just fill in the index itself.

calvinrp: Reworking and pushing shortly. Agree and see how to simplify.

calvinrp: Just pushed

PR #195, opened 2023-08-30
publish record endpoint changes contentSources to optional JSON field

calvinrp:
Currently, requires `contentSources` JSON field to be provided even if it is empty. This makes it optional and matches the OpenAPI spec.

### DIFF
--- a/crates/api/src/v1/package.rs
+++ b/crates/api/src/v1/package.rs
@@ -52,7 +52,7 @@ pub struct PublishRecordRequest<'a> {
     /// The complete set of content sources for the record.
     ///
     /// A registry may not support specifying content sources directly.
-    pub content_sources: HashMap<AnyHash, Vec<ContentSource>>,
+    pub content_sources: Option<HashMap<AnyHash, Vec<ContentSource>>>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/api/src/v1/package.rs
```diff
@@ -52,7 +52,7 @@ pub struct PublishRecordRequest<'a> {
     /// The complete set of content sources for the record.
     ///
     /// A registry may not support specifying content sources directly.
-    pub content_sources: HashMap<AnyHash, Vec<ContentSource>>,
+    pub content_sources: Option<HashMap<AnyHash, Vec<ContentSource>>>,
```
lann, line 55: If you want it to be optional for serialization purposes you can annotate it for serde:
```rust
#[serde(default, skip_serializing_if = "HashMap::is_empty")]
```

(`default` means "fill in with `Default` if missing)

calvinrp: Ah, good call. I'll revise. Thanks.

lann:
This is a breaking API change since clients will stop sending the field while the server will continue to require it. Does it need to merge soon?

lann (approved):

PR #200, opened 2023-09-25
added missing serde flatten

calvinrp:
This fixes a mismatch between the intended JSON schema, described in the [OpenAPI spec](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/bytecodealliance/registry/main/crates/server/openapi.yaml#tag/package/operation/getPackageRecord), and the implementation.

### DIFF
--- a/crates/api/src/v1/package.rs
+++ b/crates/api/src/v1/package.rs
@@ -62,6 +62,7 @@ pub struct PackageRecord {
     /// The identifier of the package record.
     pub id: RecordId,
     /// The current state of the package.
+    #[serde(flatten)]
     pub state: PackageRecordState,
 }
 


lann:
If we have other breaking protocol changes coming down the pipe lets wait to merge this to avoid needing another client/server upgrade sync.

calvinrp:
> If we have other breaking protocol changes coming down the pipe lets wait to merge this to avoid needing another client/server upgrade sync.

Yep, do you think I should combine all the changes in the same PR?

esoterra:
We could also put them on main but not cut a release or retarget this PR to a release candidate branch and do the same for other  proposed changes (maybe even make it a 0.2 branch?).

calvinrp:
Now that we resolved the release process, should we approve and merge this PR? Or should it be lumped into an upcoming PR that has other unrelated breaking changes.

esoterra:
I'm not familiar with the context for this flatten, so I'll let Lann review, but I think now that we've got the process sorted this shouldn't be considered blocked

lann (approved):

PR #201, opened 2023-09-26
Bump version 0.2.0

lann:


esoterra:
Thoughts on the release branch approach and instead taking a new branch and tagging it `0.2.0`?

lann:
I'm wary of any approach that doesn't have the latest code landing on `main`, at least for public projects. I think it makes sense to bump the version like this in any case.

calvinrp (approved):

esoterra:
When we end up working on updates to 0.2 and preparing for 0.3 at the same time, which one do you want to be on main? I was thinking it would be 0.2 on main with release candidate 0.3 on branch and that that pattern is worth starting now.

lann:
I think it's less confusing to keep the latest changes on `main` and a `release-0.2` branch for any backports.

esoterra:
If we are going to work on main with the release candidate work, how do we want to indicate when we're actually blessing 0.2?

lann:
By tagging a `v0.2.0`? We could also make it `0.2.0-pre` or something now if that seems less confusing.

esoterra:
It still feels weird to main to say `0.2` before we actually get there, but I won't fight it. The bigger thing I really want is a more structured release process that addresses how we officially bless releases, update release notes, and publish to crates.io

Just seeing PRs get thrown up to bump versions without a clear plan threw me off.

lann:
Seems like a pretty common practice: https://github.com/bytecodealliance/wasmtime/commit/e95c8556d632e32b714a97dca77ac3f4740a6e51

lann:
Sure, I think a release process is reasonable. I've been thinking of 0.2 as our first "real" release; _this_ step is basically staging for that release.

lann:
I've updated the version here to `0.2.0-dev` which is hopefully more clear.

I think version/branch-wise the release process could look something like:

- Cut a `release-0.2` branch from `main`
- Update `main` `Cargo.toml` `version = "0.3.0-dev"`
- In the `release-0.2` branch:
  - Finalize `0.2.0` release:
    - Feature freeze / bug fix period (if desired)
    - Update changelog (if we want one in-tree)
  - PR/merge `Cargo.toml` `version = "0.2.0"`
  - Tag that commit as `v0.2.0`

esoterra (approved):
That process sounds good to me

PR #202, opened 2023-09-26
changes to make it easier for other server implementations and componentizing

calvinrp:
I'm holding on the Wasm Componentizing PR until it is cleaned up.

I found the following changes make it easier for componentizing pieces of the Registry implementation as well as building another server implementation that imports from these crates.

It would be helpful to have them in the main repo as opposed to on a branch. Please let me know if some variation of this PR is acceptable to merge.

### DIFF
--- a/crates/protocol/src/operator/mod.rs
+++ b/crates/protocol/src/operator/mod.rs
@@ -9,7 +9,7 @@ use crate::{pbjson_to_prost_timestamp, prost_to_pbjson_timestamp, registry::Reco
 mod model;
 mod state;
 
-pub use model::{OperatorEntry, OperatorRecord};
+pub use model::{OperatorEntry, OperatorRecord, Permission};
 pub use state::{LogState, ValidationError};
 
 /// The currently supported operator protocol version.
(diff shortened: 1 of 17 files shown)


crates/protocol/src/operator/mod.rs
```diff
@@ -114,9 +114,10 @@ struct PermissionParseError {
 }
 
 // Serialization
+pub const SIGNING_PREFIX: &[u8] = b"WARG-OPERATOR-RECORD-SIGNATURE-V0";
```
lann, line 117: Is this to allow signing with another crypto implementation? I'd rather the interface be "the entire thing to be signed" than reimplement the slightly-subtle payload encoding.

calvinrp: This was for the "Componentizing" implementation, exposing the signing prefix outside the crate so that it is possible for the signing to happen outside of the Wasm binary. We could it as you suggest but it would involve more copying and allocations.

I'm happy to pull out these changes for the "Componentizing" implementations into a separate PR. That might make more sense. So that they can be evaluated for the full usage.

lann: These payloads are really quite small. Compared to the ECDSA signing you're doing immediately after, copying the payload out is going to be effectively free.

lann (commented):
Its hard to evaluate these changes in isolation. There is a good bit of relaxing constraints in here like making `ProtoEnvelope`'s fields `pub` that makes it harder to maintain its invariants, which makes me particularly nervous around crypto code.

esoterra:
I agree, plus there's also a general loss of type-safety with making things take bytes instead of records, which increases the chance of errors being made.

calvinrp:
Yes, I understand the concerns. Let me revise and separate in two PRs, one for the changes required to compile another server implementation and another one for componentizing the Registry logic. It will be easier to evaluate.

PR #203, opened 2023-09-27
fetch logs endpoint using fetch token String

calvinrp:
This changes the `POST /v1/fetch/logs` endpoint to use a server implementation defined `fetchToken` for the last known record instead of the `RecordId`. It still can be the `RecordId`, but it doesn't need to be. In fact, the reference implementation still uses the `RecordId` for the `fetchToken`.

The motivation here is to give server implementations greater flexibility on implementing the fetch logs endpoint.

Also, in this PR, the Client CLI will be more resilient in handling records that are returned that it has already seen before.

### DIFF
--- a/crates/api/src/v1/fetch.rs
+++ b/crates/api/src/v1/fetch.rs
@@ -6,9 +6,20 @@ use std::{borrow::Cow, collections::HashMap};
 use thiserror::Error;
 use warg_crypto::hash::AnyHash;
 use warg_protocol::{
-    registry::{LogId, RecordId, RegistryLen},
+    registry::{LogId, RegistryLen},
     PublishedProtoEnvelopeBody,
 };
+
+/// Wraps the PublishedProtoEnvelopeBody with a fetch token.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PublishedRecord {
+    /// Record proto envelope body with RegistryIndex.
+    #[serde(flatten)]
+    pub envelope: PublishedProtoEnvelopeBody,
+    /// Fetch token for fetch pagination.
+    pub fetch_token: String,
+}
 
 /// Represents a fetch logs request.
 #[derive(Debug, Serialize, Deserialize)]
@@ -19,12 +30,12 @@ pub struct FetchLogsRequest<'a> {
     /// The limit for the number of operator and package records to fetch.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub limit: Option<u16>,
-    /// The last known operator record.
+    /// The last known operator record fetch token.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub operator: Option<Cow<'a, RecordId>>,
-    /// The map of package identifiers to last known record ids.
+    pub operator: Option<Cow<'a, str>>,
+    /// The map of package identifiers to last known fetch token.
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
-    pub packages: Cow<'a, HashMap<LogId, Option<RecordId>>>,
+    pub packages: Cow<'a, HashMap<LogId, Option<String>>>,
 }
 
 /// Represents a fetch logs response.
@@ -36,10 +47,10 @@ pub struct FetchLogsResponse {
     pub more: bool,
     /// The operator records appended since the last known operator record.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub operator: Vec<PublishedProtoEnvelopeBody>,
+    pub operator: Vec<PublishedRecord>,
     /// The package records appended since last known package record ids.
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
-    pub packages: HashMap<LogId, Vec<PublishedProtoEnvelopeBody>>,
+    pub packages: HashMap<LogId, Vec<PublishedRecord>>,
 }
 
 /// Represents a fetch API error.
@@ -53,8 +64,8 @@ pub enum FetchError {
     #[error("log `{0}` was not found")]
     LogNotFound(LogId),
     /// The provided record was not found.
-    #[error("record `{0}` was not found")]
-    RecordNotFound(RecordId),
+    #[error("record fetch token `{0}` was not found")]
+    RecordNotFound(String),
     /// An error with a message occurred.
     #[error("{message}")]
     Message {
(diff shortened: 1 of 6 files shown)


crates/api/src/v1/fetch.rs
```diff
@@ -53,8 +64,8 @@ pub enum FetchError {
     #[error("log `{0}` was not found")]
     LogNotFound(LogId),
     /// The provided record was not found.
-    #[error("record `{0}` was not found")]
-    RecordNotFound(RecordId),
+    #[error("record fetch token `{0}` was not found")]
+    RecordNotFound(String),
```
lann, line 68: Maybe `InvalidFetchToken`? Also do we even need this? Now that the client can handle it we might just want to return the entire log on invalid fetch token to allow changes to the token format.

crates/api/src/v1/fetch.rs
```diff
@@ -6,10 +6,21 @@ use std::{borrow::Cow, collections::HashMap};
...
 use warg_protocol::{
-    registry::{LogId, RecordId, RegistryLen},
+    registry::{LogId, RegistryLen},
     PublishedProtoEnvelopeBody,
 };
 
+/// Wraps the PublishedProtoEnvelopeBody with a fetch token.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PublishedRecord {
```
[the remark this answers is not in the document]
calvinrp: Conceptually, I thought `fetchToken` is part of the `api` crate and not the `protocol` crate (where the `PublishedProtoEnvelopeBody` lives). Also, the `PublishedProtoEnvelopeBody` is used for the `GET /v1/package/{logId}/record/{recordId}` endpoint. The `fetchToken` is only relevant for the `POST /v1/fetch/logs`.

Thought a wrapper made the most sense. But let me know your thoughts on this.

crates/api/src/v1/fetch.rs
calvinrp: I think an error makes sense. Happy to rename it to `InvalidFetchToken`.

lann: How would a server migrate fetch token formats?

calvinrp: With this PR, as is:

TLDR: it would work just fine.

The server implementation isn't required to return an error for `InvalidFetchToken`. If it does not error and returns more records than necessary, the client will ignore the records that it already has (without updating to the new `fetchToken` for the previous records). Once a new record is encountered, that new `fetchToken` is now saved and used for future pagination.

I would assume that a server implementation that is migrating to a new `fetchToken` format would not error and would still support the old format (detecting the difference if format versions) for some period of time.

calvinrp: I'll see if I can push a change that makes the client more resilient for `fetchToken` migration.

calvinrp: We can discuss on the Registry call, but this implementation is likely sufficiently resilient as is.

calvinrp: Renamed the `RecordNotFound` for the `FetchError` to `FetchTokenNotFound`. Hope that feels better.

crates/api/src/v1/fetch.rs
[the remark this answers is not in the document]
lann: Not I big deal for now. Just a lot of layers of indirection at this point.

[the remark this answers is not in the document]
calvinrp: Yeah, agree. Probably ok for now.

crates/client/src/lib.rs
```diff
@@ -405,12 +399,21 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
-                    .validate(&record.envelope)
-                    .map_err(|inner| ClientError::OperatorValidationFailed { inner })?;
-                operator.head_registry_index = Some(record.registry_index);
+                let proto_envelope: PublishedProtoEnvelope<operator::OperatorRecord> =
+                    record.envelope.try_into()?;
+                if operator.head_registry_index.is_none()
+                    || operator.head_registry_index.is_some_and(|registry_index| {
+                        proto_envelope.registry_index > registry_index
+                    })
+                {
```
lann, line 408: Alternative wording:

```suggestion
                if proto_envelope.registry_index > operator.head_registry_index.unwrap_or(0) {
```

crates/client/src/lib.rs
```diff
@@ -405,12 +399,21 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+                    // skips over records that has already seen
```
lann, line 409: Move outside of the `if`?

crates/client/src/lib.rs
```diff
@@ -419,15 +422,24 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
-                    })?;
-                    package.head_registry_index = Some(record.registry_index);
+                    let proto_envelope: PublishedProtoEnvelope<package::PackageRecord> =
+                        record.envelope.try_into()?;
+                    if package.head_registry_index.is_none()
+                        || package.head_registry_index.is_some_and(|registry_index| {
+                            proto_envelope.registry_index > registry_index
+                        })
+                    {
+                        // skips over records that has already seen
```
lann, line 432: Ditto above.

crates/server/openapi.yaml
```diff
@@ -802,17 +805,22 @@ components:
...
             - registryIndex
+            - fetchToken
           properties:
             registryIndex:
               type: integer
               description: The index of the published record in the registry log.
               example: 42
+            fetchToken:
+              type: string
+              description: The fetch token for the registry log. Used for pagination in the fetch logs endpoint.
```
lann, line 822: :shrug: 
```suggestion
              description: The fetch token for the registry log. Used for incremental sync in the fetch logs endpoint.
```

lann (approved):
A few style/wording nits. Nothing blocking.

crates/server/openapi.yaml
calvinrp: `Used as a cursor for incrementally fetching log records.` ?

crates/client/src/lib.rs
calvinrp: Thanks for the suggestion, still getting better at Rust. I'll fix. Doesn't quite work I think given it is an unsigned integer and `0` is a valid (and likely registry index for the Operator log). Seeing if it can be cleaner.

calvinrp: So I can simplifiy it, to `operator.head_registry_index.is_none() || proto_envelope.registry_index > operator.head_registry_index.unwrap()`. Since the `.is_none() || ` will ensure the `.unwrap()` won't panic.

lann (approved):

PR #212, opened 2023-09-28
Added protobuf for stacklog and veclog and a few methods helpful for other implementations

calvinrp:
Added protobuf for stacklog and veclog, which isn't used by the reference implementation at the moment but is helpful for other implementations;

Added:
- `AnyHash::new()`
- `impl From<RecordId> for AnyHash` and `impl AsRef<[u8]> for RecordId`

Changed:
- `from_protobuf(bytes: Vec<u8>)` to `from_protobuf(bytes: &[u8])` for `ProtoEnvelope`


crates/protocol/src/registry.rs
```diff
@@ -161,6 +161,11 @@ impl PackageId {
     pub fn name(&self) -> &str {
         &self.id[self.colon + 1..]
     }
+
+    /// Gets the full string package identifier.
+    pub fn id(&self) -> &str {
+        &self.id
+    }
```
lann, line 168: Do we need this in addition to the `AsRef<str>`?

crates/transparency/src/log/vec_log.rs
```diff
@@ -67,6 +70,19 @@ where
 
         Some(result)
     }
+
+    /// Turn a VecLog into bytes using protobuf
+    pub fn encode(self) -> Vec<u8> {
```
lann, line 75: I would call these `to_/from_protobuf` for consistency with ProtoEnvelope.

proto/warg/transparency/proofs.proto
```diff
@@ -24,4 +24,14 @@ message LogProofBundle {
 message HashEntry {
     uint32 index = 1;
     bytes hash = 2;
-}
\ No newline at end of file
+}
+
+message StackLog {
```
lann, line 29: Are we even using StackLog any more? I don't see any references to it outside of `warg-transparency`...

crates/protocol/src/registry.rs
calvinrp: Nope, I missed that. I'll remove.

proto/warg/transparency/proofs.proto
calvinrp: Yes, we use the `StackLog` for computing the `logRoot` on Checkpoints. `VecLog` is used for computing the log proof bundles.

crates/transparency/src/log/vec_log.rs
calvinrp: Cool with me. I'll rename.

proto/warg/transparency/proofs.proto
lann: Is that happening somewhere in this repo? I just don't see any references to `StackLog` outside of this crate...

lann: It looks like *I* removed it in https://github.com/bytecodealliance/registry/pull/169 though I have no recollection of that :sweat_smile:

calvinrp: Yes, looks like you may be correct. Looks like the `CoreService` uses the `VecLog` for computing the `logRoot`. Both `StackLog` can compute the `logRoot`. From my understanding, the `StackLog` memory usage scales differently than the `VecLog`.

There's an argument to be made for dropping unused code. I'm currently using both in my server implementation. It would be helpful to keep both around and revisit this.

lann: OK - part of the reason I was asking is that it feels wrong to expose otherwise-hidden implementation details here where we are otherwise defining a public interface to the service. It might be worthwhile to put these messages in a separate file (`log_impls.proto`?) and hide the generated code behind a feature flag. If that is too big a lift right now I think comments on the message types here to the effect of "these are internal implementation details of the server, not part of the client contract" would be OK in the medium-term.

calvinrp: Understand. I'll look into it.

calvinrp: Pushed an update. Created an `proto/warg/internal/internal.proto`. That probably is enough and doesn't need to be behind a feature flag, but open to doing that. Let me know.

lann (approved):

PR #216, opened 2023-10-15
Proposed changes before release

calvinrp:
Proposed breaking changes to the API:
1. `GET /v1/content/{digest}` endpoint instead of `GET /v1/package/{logId}/record/{recordId}` for getting content source download URLs;
  Current API is not ideal, since records can have multiple releases which provides more download links than clients likely want. Also, when Wasm naked imports are implemented, this new endpoint could more cleanly be extended to provide the dependency URLs for each digest.
2. Upload HTTP URLs can now specific the `method` and `headers`, so that can use `POST` or `PUT` as well as `Authorization` and `Content-Type` headers; In the future, it would be helpful to add multipart upload support for large files as a new variant type;
3. Download HTTP URLs can specify whether the `Range` header is supported to download pieces of large files in parallel; Optional `sizeHint` field is provided to hint at the expected file size; (Although client doesn't yet take advantage of this functionality.)

Proposed additive changes:
4. Added `warg reset` command to reset local registry state (clearing package logs etc);
5. Added `warg clear` command to clear local cache of downloaded content;

Also, there's a fix so that the client CLI verifies the `TimestampedCheckpoint` signature when updating checkpoints. I'm pretty sure it was not verifying the signature before.

### DIFF
--- a/crates/api/src/v1/content.rs
+++ b/crates/api/src/v1/content.rs
@@ -1,0 +1,109 @@
+//! Types relating to the content API.
+
+pub use super::ContentSource;
+use crate::Status;
+use serde::{de::Unexpected, Deserialize, Serialize, Serializer};
+use std::{borrow::Cow, collections::HashMap};
+use thiserror::Error;
+use warg_crypto::hash::AnyHash;
+
+/// Represents a response for content digest.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ContentSourcesResponse {
+    /// The content sources for the requested content digest, as well as additional
+    /// content digests imported by the requested content digest.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub content_sources: HashMap<AnyHash, Vec<ContentSource>>,
+}
+
+/// Represents a content API error.
+#[non_exhaustive]
+#[derive(Debug, Error)]
+pub enum ContentError {
+    /// The provided content digest was not found.
+    #[error("content digest `{0}` was not found")]
+    ContentDigestNotFound(AnyHash),
+    /// An error with a message occurred.
+    #[error("{message}")]
+    Message {
+        /// The HTTP status code.
+        status: u16,
+        /// The error message
+        message: String,
+    },
+}
+
+impl ContentError {
+    /// Returns the HTTP status code of the error.
+    pub fn status(&self) -> u16 {
+        match self {
+            Self::ContentDigestNotFound(_) => 404,
+            Self::Message { status, .. } => *status,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+enum EntityType {
+    ContentDigest,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(untagged, rename_all = "camelCase")]
+enum RawError<'a, T>
+where
+    T: Clone + ToOwned,
+    <T as ToOwned>::Owned: Serialize + for<'b> Deserialize<'b>,
+{
+    NotFound {
+        status: Status<404>,
+        #[serde(rename = "type")]
+        ty: EntityType,
+        id: Cow<'a, T>,
+    },
+    Message {
+        status: u16,
+        message: Cow<'a, str>,
+    },
+}
+
+impl Serialize for ContentError {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        match self {
+            Self::ContentDigestNotFound(digest) => RawError::NotFound {
+                status: Status::<404>,
+                ty: EntityType::ContentDigest,
+                id: Cow::Borrowed(digest),
+            }
+            .serialize(serializer),
+            Self::Message { status, message } => RawError::Message::<()> {
+                status: *status,
+                message: Cow::Borrowed(message),
+            }
+            .serialize(serializer),
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for ContentError {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        match RawError::<String>::deserialize(deserializer)? {
+            RawError::NotFound { status: _, ty, id } => match ty {
+                EntityType::ContentDigest => Ok(Self::ContentDigestNotFound(
+                    id.parse::<AnyHash>().map_err(|_| {
+                        serde::de::Error::invalid_value(Unexpected::Str(&id), &"a valid digest")
+                    })?,
+                )),
+            },
+            RawError::Message { status, message } => Ok(Self::Message {
+                status,
+                message: message.into_owned(),
+            }),
+        }
+    }
+}
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -19,9 +19,9 @@ pub enum ContentSource {
         /// Optional support for HTTP Range header.
         #[serde(default, skip_serializing_if = "is_false")]
         supports_range_header: bool,
-        /// Optional content size hint in bytes.
+        /// Optional content size in bytes.
         #[serde(skip_serializing_if = "Option::is_none")]
-        size_hint: Option<u64>,
+        size: Option<u64>,
     },
 }
 
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -7,7 +7,6 @@ pub mod paths;
 pub mod proof;
 
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
 
 /// Represents the supported kinds of content sources.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -17,9 +16,6 @@ pub enum ContentSource {
     HttpGet {
         /// The URL of the content.
         url: String,
-        /// Optional header names and values for the request.
-        #[serde(default, skip_serializing_if = "HashMap::is_empty")]
-        headers: HashMap<String, String>,
         /// Optional support for HTTP Range header.
         #[serde(default, skip_serializing_if = "is_false")]
         supports_range_header: bool,
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -16,9 +16,9 @@ pub enum ContentSource {
     HttpGet {
         /// The URL of the content.
         url: String,
-        /// Optional support for HTTP Range header.
+        /// Optional accepts for HTTP Range header.
         #[serde(default, skip_serializing_if = "is_false")]
-        supports_range_header: bool,
+        accept_ranges: bool,
         /// Optional content size in bytes.
         #[serde(skip_serializing_if = "Option::is_none")]
         size: Option<u64>,
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -1,6 +1,34 @@
 //! Types representing v1 of the Warg REST API.
 
+pub mod content;
 pub mod fetch;
 pub mod package;
 pub mod paths;
 pub mod proof;
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// Represents the supported kinds of content sources.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "camelCase")]
+pub enum ContentSource {
+    /// The content is located at an GET HTTP URL.
+    HttpGet {
+        /// The URL of the content.
+        url: String,
+        /// Optional header names and values for the request.
+        #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+        headers: HashMap<String, String>,
+        /// Optional support for HTTP Range header.
+        #[serde(default, skip_serializing_if = "is_false")]
+        supports_range_header: bool,
+        /// Optional content size in bytes.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        size: Option<u64>,
+    },
+}
+
+fn is_false(b: &bool) -> bool {
+    !b
+}
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -16,10 +16,10 @@ pub enum ContentSource {
     HttpGet {
         /// The URL of the content.
         url: String,
-        /// Optional accepts for HTTP Range header.
+        /// Optional, server accepts for HTTP Range header.
         #[serde(default, skip_serializing_if = "is_false")]
         accept_ranges: bool,
-        /// Optional content size in bytes.
+        /// Optional, provides content size in bytes.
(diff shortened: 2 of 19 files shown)


lann:
> GET /v1/content/{digest} endpoint instead of GET /v1/package/{logId}/record/{recordId} for getting content source download URLs;
Current API is not ideal, since records can have multiple releases which provides more download links than clients likely want. Also, when Wasm naked imports are implemented, this new endpoint could more cleanly be extended to provide the dependency URLs for each digest.

My only concern here is that its an extra round-trip for the common flow of fetching/validating a package log immediately before fetching contents.

lann:
> Upload HTTP URLs can now specific the method and headers, so that can use POST or PUT and possibly Authorization or other headers; In the future, it would be helpful to add multipart upload support for large files;

I think this is probably fine, but my intention was to add `UploadEndpoint` variants if this gets much more complicated.

lann:
> Download HTTP URLs can specify headers and whether the Range header is supported to download pieces of large files in parallel; Optional size field is provided to hint at the expected file size; (Although client doesn't yet take advantage of this functionality.)

I'm somewhat wary of adding "headers" here specifically. I'd really like this baseline download URL type to work with _just_ the URL to make sure it works in as many client scenarios as possible. Do you have a use case where headers are required on download?

calvinrp:
> > GET /v1/content/{digest} endpoint instead of GET /v1/package/{logId}/record/{recordId} for getting content source download URLs;
> > Current API is not ideal, since records can have multiple releases which provides more download links than clients likely want. Also, when Wasm naked imports are implemented, this new endpoint could more cleanly be extended to provide the dependency URLs for each digest.
> 
> My only concern here is that its an extra round-trip for the common flow of fetching/validating a package log immediately before fetching contents.

At moment, this endpoint `GET /v1/package/{logId}/record/{recordId}` is being used for 2 use cases:
1. polling for publish status (`sourcing`, `processing`, `rejected`, `published`) after a publish attempt;
2. requesting download URLs for content;

This proposed change separates these concerns into different endpoints.

At the moment, this current endpoint is not used for the fetching/validating step, but rather only the `POST /v1/fetch/logs` endpoint. The `record` base64-encoded protobuf returned when `published` status was not being used by the client, since the client already had it when attempted to publish.

In addition, each record could have many release entries and the client often only cares about a single release. The current endpoint would return download URLs for all release entries published in a record.

calvinrp:
> > Upload HTTP URLs can now specific the method and headers, so that can use POST or PUT and possibly Authorization or other headers; In the future, it would be helpful to add multipart upload support for large files;
> 
> I think this is probably fine, but my intention was to add `UploadEndpoint` variants if this gets much more complicated.

Yep, that keeps this pattern to extend the `enum` variants. It just adds a bit more flexibility in what HTTP requests can be made, without adding much complexity.

I was thinking that multipart uploads would be another variant type that could be used for large file uploads.

calvinrp:
> > Download HTTP URLs can specify headers and whether the Range header is supported to download pieces of large files in parallel; Optional size field is provided to hint at the expected file size; (Although client doesn't yet take advantage of this functionality.)
> 
> I'm somewhat wary of adding "headers" here specifically. I'd really like this baseline download URL type to work with _just_ the URL to make sure it works in as many client scenarios as possible. Do you have a use case where headers are required on download?

The ability to provide a `Authorization` header does come in handy.

lann:
> The ability to provide a Authorization header does come in handy.

True, but its also an easy place for a client implementation to trip up and e.g. leak credentials. I'm just trying to understand whether this needs to be part of the "baseline" http download type.

lann:
> Is the concern that it is too much flexibility in what headers that could be provided?

Yes, because some clients (browser `fetch`, notably) can't send all headers in all contexts, e.g. under CORS constraints.

calvinrp:
> > Is the concern that it is too much flexibility in what headers that could be provided?
> 
> Yes, because some clients (browser `fetch`, notably) can't send all headers in all contexts, e.g. under CORS constraints.

Yeah, agree with those considerations. If it was a protocol restriction as opposed to a registry implementation's choice, that does affect usage patterns.

Not sure if it changes things or not, but it is worth pointing out that the endpoint returns an list (`JSON Array`) of download options. In theory, there could be options. But not sure if that makes sense in this particular case.

I'll drop the `headers` field for download URLs. Can choose to add later.

crates/protocol/src/registry.rs
```diff
@@ -161,6 +161,16 @@ impl PackageId {
     pub fn name(&self) -> &str {
         &self.id[self.colon + 1..]
     }
+
+    /// Check if string is a valid namespace.
+    pub fn is_valid_namespace(namespace: &str) -> bool {
+        const SUPPORTS_NESTED_NAMESPACES: bool = false;
```
esoterra, line 167: What's this about?

calvinrp: From my understanding, `gh:user:project` nested namespaces will be eventually supported when the tooling catches up. Wanted to make it easy to support but not enable yet. Can remove it or refactor if you have a better convention.

crates/api/src/v1/mod.rs
```diff
@@ -1,6 +1,30 @@
...
+    /// The content is located at an GET HTTP URL.
+    HttpGet {
+        /// The URL of the content.
+        url: String,
+        /// Optional support for HTTP Range header.
+        #[serde(default, skip_serializing_if = "is_false")]
+        supports_range_header: bool,
+        /// Optional content size in bytes.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        size: Option<u64>,
```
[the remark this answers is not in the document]
calvinrp: Ok with me, more explicit.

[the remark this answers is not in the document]
lann: `size_hint` implies (to me) that it is an upper or lower bound, not an exact size.

crates/api/src/v1/mod.rs
```diff
@@ -1,6 +1,30 @@
...
+/// Represents the supported kinds of content sources.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "camelCase")]
+pub enum ContentSource {
+    /// The content is located at an GET HTTP URL.
+    HttpGet {
+        /// The URL of the content.
+        url: String,
+        /// Optional support for HTTP Range header.
+        #[serde(default, skip_serializing_if = "is_false")]
```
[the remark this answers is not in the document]
calvinrp: Does not appear to work. `error[E0277]: the trait bound `ContentSource: Serialize` is not satisfied`

crates/protocol/src/registry.rs
calvinrp: I knew the compiler would strip it out at compile-time. Didn't think it merited a build flag. And wanted to have the solution easy to enable in the code.

But I'll push a change to remove it and only validate non-nested namespaces.

calvinrp: Removed

calvinrp:
Keep command name for `warg reset`, but rename `warg clear` to something else?

crates/api/src/v1/mod.rs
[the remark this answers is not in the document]
calvinrp: In practice, `size` should be the exact size. However, it isn't part of any verifiable data. I changed to `size_hint` but I can change back or rename it to something else. Let me know and I'll update the PR.

The intention of `supports_range_header` and `size` is giving the client enough info to determine if it should try to download in concurrent chunks instead of all at once.

Alternatively, the server could make the recommendation on whether to use the `Range` header and how many chunks (and what size). But I'm not sure I like that approach.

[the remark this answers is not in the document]
calvinrp: I personally prefer, `size` over `size_hint`.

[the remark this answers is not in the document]
lann: I don't think it needs to be verifiable to be useful, but I do think we should specify whether or not it needs to be exact (if present).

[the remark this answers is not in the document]
calvinrp: I pushed a change, renaming back to `size`.

lann:
Could also combine the functionality of `warg reset` and `warg clear` with an opt-in flag e.g. `warg reset --clear-content`.

lann:
I think I would also slightly favor `warg clear` over `warg reset`, but that might just be my Rust (`cargo clear`) bias.

calvinrp:
> Could also combine the functionality of `warg reset` and `warg clear` with an opt-in flag e.g. `warg reset --clear-content`.

Combining the commands like so wouldn't give you the ability to clear the content cache without clearing the registry log data. Not sure if that matters.

calvinrp:
> I think I would also slightly favor `warg clear` over `warg reset`, but that might just be my Rust (`cargo clear`) bias.

I was thinking of `reset` because we were referring to doing a "trust reset", which would require clearing previous registry log data and requesting the updated logs. However, I don't feel strongly about the command names. Just need something that does the job in the CLI without manually doing it (annoying).

crates/api/src/v1/mod.rs
```diff
@@ -1,6 +1,30 @@
...
+        supports_range_header: bool,
```
lann, line 21: Bike-shedding the name; how about `accept_ranges`, mirroring the roughly-equivalent `Accept-Ranges` HTTP header?

crates/api/src/v1/package.rs
```diff
@@ -10,27 +11,21 @@ use warg_protocol::{
...
 #[serde(tag = "type", rename_all = "camelCase")]
 pub enum UploadEndpoint {
-    /// Content may be uploaded via HTTP POST to the given URL.
-    /// If the endpoint responds with "201 Created" and a Location header, that
-    /// header's value will be the content source.
-    HttpPost {
+    /// Content may be uploaded via HTTP request to the given URL.
+    Http {
+        /// The http method for the upload request.
+        /// Only `POST` and `PUT` methods are accepted by the client.
```
lann, line 21: This reads slightly more clearly to me:

```suggestion
        /// Only `POST` and `PUT` methods are supported.
```

crates/api/src/v1/package.rs
```diff
@@ -10,27 +11,21 @@ use warg_protocol::{
...
+        method: String,
         /// The URL to POST content to.
         url: String,
+        /// Optional header names and values for the upload request.
+        /// Only `authorization` and `content-type` headers are accepted by the client.
```
lann, line 26: I think we should be stronger with this language. I think it's plausible that someone else would implement a partial client which would probably include this endpoint.

```suggestion
        /// Only `authorization` and `content-type` headers are valid; any other header should be rejected.
```

calvinrp:
It probably isn't that big of deal to rename the CLI commands (`reset` and `clear`) later. I doubt there will be many people initially using these commands regularly (except for Warg devs). Not a lot of muscle memory to fight after a rename.

That said, happy to reconfigure it in this PR. Just let me know preferences.

crates/api/src/v1/package.rs
```diff
@@ -58,9 +53,10 @@ pub struct PublishRecordRequest<'a> {
 
 /// Represents a package record API entity in a registry.
 #[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct PackageRecord {
     /// The identifier of the package record.
-    pub id: RecordId,
+    pub record_id: RecordId,
```
lann, line 59: I don't think we discussed this change; I'm not opposed to it but I'd like to be consistent across APIs, e.g. changing package IDs to `package_id` as well.

crates/api/src/v1/mod.rs
```diff
@@ -1,6 +1,30 @@
...
 pub mod paths;
 pub mod proof;
+
+use serde::{Deserialize, Serialize};
+
+/// Represents the supported kinds of content sources.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "camelCase")]
+pub enum ContentSource {
+    /// The content is located at an GET HTTP URL.
```
lann, line 15: :shrug: 
```suggestion
    /// The content can be retrieved with an HTTP GET.
```

crates/server/openapi.yaml
```diff
@@ -216,54 +210,35 @@ paths:
...
-            schema:
-              type: string
-              format: binary
-              maxLength: 1073741824 # 1 GiB limit
+        - name: X-Proxy-Registry
+          in: header
+          description: The proxy registry URI Authority context to evaluate the request instead of the current registry.
+          required: false
+          schema:
+            type: string
```
lann, line 234: Out of scope for this PR?

lann:
> It probably isn't that big of deal to rename the CLI commands (reset and clear) later.

Agreed

> I was thinking of reset because we were referring to doing a "trust reset"

Out of scope for this PR, but I have imagined a "trust reset" to be an entirely separate UX flow where the implications of the trust reset are explained to the user before they opt-in.

crates/api/src/v1/package.rs
calvinrp: Yeah, more explicit is better. I'll update `package_id` as well.

crates/server/openapi.yaml
calvinrp: Oops, yeah that wasn't supposed to be in there. Pulling things out of another PR. I'll fix.

crates/api/src/v1/content.rs
```diff
@@ -0,0 +1,109 @@
...
+use warg_crypto::hash::AnyHash;
+
+/// Represents a response for content digest.
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ContentSourcesResponse {
+    /// The content sources for the requested content digest, as well as additional
+    /// content digests imported by the requested content digest.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub content_sources: HashMap<AnyHash, Vec<ContentSource>>,
```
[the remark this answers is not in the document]
calvinrp: Should I try to do that now?

crates/api/src/v1/package.rs
calvinrp: Updated `package_id` as well to be more consistent throughout.

lann (approved):

crates/api/src/v1/content.rs
```diff
@@ -0,0 +1,109 @@
...
+        ty: EntityType,
+        id: Cow<'a, T>,
+    },
+    Message {
+        status: u16,
+        message: Cow<'a, str>,
+    },
+}
+
+impl Serialize for ContentError {
```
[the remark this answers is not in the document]
calvinrp: Hmm, I copied and modified this code from the existing pattern:

https://github.com/bytecodealliance/registry/blob/main/crates/api/src/v1/package.rs#L206-L288
https://github.com/bytecodealliance/registry/blob/main/crates/api/src/v1/fetch.rs#L122-L179
https://github.com/bytecodealliance/registry/blob/main/crates/api/src/v1/proof.rs#L143-L215

Do you think they should all be revised? Or fine for now?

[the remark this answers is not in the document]
lann: Totally fine; as-is is correct and even preferable if performance is a concern. Just pointing out a simpler alternative.

Issue #224, opened 2023-11-01
Add Namespace Definition/Imports to Operator Log


esoterra:
As proposed in November 1st meeting, I think a good way to start prototyping federation is to introduce a basic namespace-to-origin mapping into the registrty state via two new Operator Log Entries.

1. `ImportNamespace { name: String, registry: String }` - This registry imports the namespace `name` from `registry`.
2. `DefineNamespace { name: String }` - This registry defines the namespace `name` in its own logs.

A registry's `name` is currently required to be a domain name and a registry's "location" is currently required to map to itself. In the future we can relax this constraint if we want to add indirection between registry name and registry location.

As a validation rule, it would be illegal for an operator log to contain two or more `*Namespace` entries using the same `name` value. This means that every namespace is mapped to exactly zero or one registries (either a remote or this one) and that that mapping does not change over time.

This feature when combined with the previously discussed `Warg-Registry` header will make it possible for clients to reason about the relationships between registries and talk to registries about each other. Most (or all?) registries will want to require that releases only reference namespaces that are bound and that requests with `Warg-Registry` headers only use header values which are registries in the operator log as this acts as a record of which registries they federate with.

Future Note 1: `ImportNamespace` can be extended in the future with an additional field that indicates what name from the specified `registry` is being imported as `name` and when no value is provided for that it is implicitly the same as `name`.

Future Note 2: The no duplicate definitions/imports validation rule can be relaxed in the future in order to introduce namespace re-mapping.

calvinrp:
To clarify, `registry: String` and its use in the `Warg-Registry` header need to validate as a domain name or authority?

lann:
We could add a header to federated (`Warg-Registry`) responses:
`Warg-Registry-Location: <registry URL>` 

In the short-term that would always be e.g. `https://{registry}` but it might let us avoid needing to require that `registry` must be a domain name. In the future (with more design work) it might also be a good mechanism for updating an imported registry's origin URL.

calvinrp:
To elaborate...

`Warg-Registry` header and `registry` values in the proposed operator log entry are arbitrary unique string IDs (but not necessarily globally unique across other registries).

To handle the use case where a client wants to check the actual source registry API instead of what is provided by the proxying registry, a `Warg-Registry-Location` header is _optionally_ provided with the current URL to the source registry. That source registry URL may change over time but the identifier used in the `Warg-Registry` header wouldn't change.

How do we feel about that? 

[...]

PR #225, opened 2023-11-08
Monitors and federation server side

calvinrp:
Implements the server-side for the following issues:
- #224
- #227
- #228
- #232 

Deferred #229 for a follow up PR, due to the large diff.


calvinrp:
In progress:
- adding tests
- validation on package publish for operator log defined namespaces

But helpful to start reviewing...

crates/protocol/src/operator/state.rs
```diff
@@ -334,6 +372,25 @@ impl LogState {
...
     }
 
+    fn validate_namespace(
+        &mut self,
+        name: &str,
+        definition: NamespaceDefinition,
+    ) -> Result<(), ValidationError> {
+        let name = name.to_lowercase();
+
+        if !PackageId::is_valid_namespace(&name) {
```
esoterra, line 382: Check that name is lowercase in this function

calvinrp: Waiting for the outcome of the [discussion on package names and namespaces](https://github.com/bytecodealliance/registry/discussions/231)

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// The provided checkpoint log length is greater than the most recent checkpoint. The monitor
+    /// is expected to only return this after it is certain that it has the most recent checkpoint.
+    #[error("checkpoint log length `{0} is greater than the most recent checkpoint")]
+    CheckpointNotFound(RegistryLen),
+    /// The checkpoint signature keyId was not found.
+    #[error("checkpoint signature keyId `{0}` is invalid for checkpoint")]
+    CheckpointSignatureKeyIdInvalid(signing::KeyID),
+    /// The checkpoint signature invalid.
+    #[error("checkpoint signature `{0}` is invalid")]
+    CheckpointSignatureInvalid(signing::Signature),
```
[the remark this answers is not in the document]
calvinrp: > If a registry doesn't feel like verifying that an old key had permission at the correct time to sign this checkpoint it can simply not verify and return Unverified.

After thinking through... If a key is revoked, it could have been compromised. We should return an error indicating that `KeyID` was revoked the permission to sign checkpoints.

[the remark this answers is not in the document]
calvinrp: Revised

crates/protocol/src/operator/state.rs
calvinrp: Modified to keep the case for package names and namespaces for presentation and code generation. But enforces uniqueness in a case insensitive way.

calvinrp:
NOTE:  This changes the `LogId` hash to always use the lowercased package name. This would be a breaking change for anything other lowercased package names.

crates/server/src/api/v1/monitor.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // verify map root, if not returns CheckpointMapRootInvalid
+    if checkpoint.as_ref().checkpoint.map_root != body.as_ref().checkpoint.map_root {
+        return Err(MonitorApiError(MonitorError::CheckpointMapRootInvalid(
+            body.as_ref().checkpoint.map_root.clone(),
+        )));
+    }
+
+    Ok(Json(CheckpointVerificationResponse::Verified))
+}
```
calvinrp, line 109: @esoterra hoping this will address your concerns on order of operation in verifying the checkpoint. Comments describe the possible validation errors can be returned at each step.

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "state", rename_all = "camelCase")]
+#[allow(clippy::large_enum_variant)]
+pub enum CheckpointVerificationResponse {
+    /// The checkpoint is verified.
+    #[serde(rename_all = "camelCase")]
+    Verified,
+    /// The checkpoint is verified but has not verified the signature.
+    #[serde(rename_all = "camelCase")]
+    VerifiedButSignatureUnverified,
```
lann, line 21: Maybe merge these?

```suggestion
    Verified {
      #[serde(skip_serializing_if = "is_false")]
      signature_verified: bool,
    }
```

Some (most?) consumers won't care whether the signature has been verified or not.

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// The checkpoint is unverified but should retry verification.
+    #[serde(rename_all = "camelCase")]
+    Retry {
```
lann, line 24: Can we use an adjective (phrase) here? Something like `notYetVerified`?

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+        /// Instructs to retry but wait the specified number of seconds.
+        wait_seconds: u16,
+    },
+    /// The checkpoint is unverified and should not retry.
+    #[serde(rename_all = "camelCase")]
+    Unverified,
```
lann, line 30: What does this state indicate? I thought it was "we haven't seen that checkpoint" but it looks like that is covered by `MonitorError::CheckpointNotFound`

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+#[non_exhaustive]
+#[derive(Debug, Error)]
+pub enum MonitorError {
+    /// The provided checkpoint log length is not known to the registry.
+    #[error("checkpoint log length `{0}` is not known to the registry")]
+    CheckpointNotFound(RegistryLen),
+    /// The provided checkpoint log length is greater than the most recent checkpoint. The monitor
+    /// is expected to only return this after it is certain that it has the most recent checkpoint.
+    #[error("checkpoint log length `{0}` is invalid and greater than the most recent checkpoint")]
+    CheckpointInvalidLogLength(RegistryLen),
```
lann, line 43: I'm not sure that this error makes sense as written. In a distributed system you never really know that you "have the most recent" of something...

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+use thiserror::Error;
+use warg_crypto::hash::AnyHash;
+use warg_crypto::signing;
+use warg_protocol::registry::RegistryLen;
+
+/// Represents checkpoint verification response in one of the following states:
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "state", rename_all = "camelCase")]
+#[allow(clippy::large_enum_variant)]
+pub enum CheckpointVerificationResponse {
```
lann, line 15: Coming back here after reading through the variants and errors below: I wonder if we'd be better off modeling everything other than "verified" as an error here (with `signature_verified: bool` merged into "verified" as suggested below). We could instead categorize errors as either "crticial" (something is positively wrong) or "non-critical" (either "maybe later" or "we don't know"). As-is both kinds of issues are sort of arbitrarily split between OK and error variants.

lann (commented):
Partial review covering the api crate changes

crates/api/src/v1/monitor.rs
calvinrp: Walked Kyle thru the PR changes again this morning. He thought it would be better if we responded with "parallel" tracks for what was verified or unverified. The checkpoint (log length, map root and log root) and the signature (key ID, signature). And just using HTTP success response codes to indicate that the request succeeded but the client needs to interpret the response body for the verification results. Thoughts?

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[serde(rename_all = "camelCase")]
+    Verified,
+    /// The checkpoint is verified but has not verified the signature.
+    #[serde(rename_all = "camelCase")]
+    VerifiedButSignatureUnverified,
+    /// The checkpoint is unverified but should retry verification.
+    #[serde(rename_all = "camelCase")]
+    Retry {
+        /// Instructs to retry but wait the specified number of seconds.
+        wait_seconds: u16,
```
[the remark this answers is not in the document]
calvinrp: Switch to the HTTP header instead of the response JSON for retry? Kind of like that.

crates/api/src/v1/monitor.rs
lann: Sure sounds fine.

crates/api/src/v1/monitor.rs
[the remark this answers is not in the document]
lann: I was thinking just the name. I wouldn't use the header unless it's paired with an http error.

crates/api/src/v1/monitor.rs
lann: How would that be different than `CheckpointNotFound`? Anyway if you're planning to change these around I'll wait to see where it ends up

crates/api/src/v1/monitor.rs
calvinrp: This is attempting to handle the case where the requested checkpoint is actually in the future. The thinking is that a monitor would first respond with `retry` and check back. And the monitor would check that it has the most recent checkpoint, and then on the next `retry` be able to answer definitely that it is not a valid `logLength`.

Let's say someone wanted to circumvent the verification process and consistently created invalid checkpoints in the future. Without some mechanism like this, the best the monitor could do is say, "I don't know". This gives the mechanism to actually return "not valid".

crates/api/src/v1/monitor.rs
calvinrp: Yeah, thinking about refactoring, as I alluded to in https://github.com/bytecodealliance/registry/pull/225#discussion_r1391489723

crates/api/src/v1/ledger.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Last registry index that is included in this ledger source.
+    pub last_registry_index: RegistryIndex,
+    /// The HTTP GET URL location for the ledger source.
+    pub url: String,
+    /// Optional, server accepts for HTTP Range header.
+    #[serde(skip_serializing_if = "is_false")]
+    pub accept_ranges: bool,
+    /// Content type for the ledger source.
+    #[serde(default, skip_serializing_if = "is_ledger_packed")]
+    pub content_type: LedgerSourceContentType,
```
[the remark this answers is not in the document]
calvinrp: Updated

crates/api/src/v1/ledger.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// Content type for the ledger source.
+#[derive(Default, PartialEq, Serialize, Deserialize)]
+pub enum LedgerSourceContentType {
+    /// The content type is binary representation of the LogId and RecordId hashes without padding.
+    /// In the case of `sha256` hash algorithm, this is a repeating sequence of 64 bytes (32 bytes
+    /// for each the LogId and RecordId) without padding.
+    #[default]
+    #[serde(rename = "application/vnd.bytecodealliance.registry.ledger.packed")]
```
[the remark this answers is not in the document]
calvinrp: Updated

crates/api/src/v1/monitor.rs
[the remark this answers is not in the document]
calvinrp: I revised to return `503` status code and use the `Retry-After` header.

crates/api/src/v1/monitor.rs
calvinrp: Refactored, but I might have misunderstood the original question.

calvinrp:
Refactored the verify checkpoint endpoint. The [OpenAPI rendered spec](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/calvinrp/bytecodealliance-registry/monitors-and-federation-server-side/crates/server/openapi.yaml#tag/monitor/operation/verifyCheckpoint)

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,124 @@
...
+    Invalid,
+}
+
+/// Represents a monitor API error.
+#[non_exhaustive]
+#[derive(Debug, Error)]
+pub enum MonitorError {
+    /// Instruct to retry after specified number of seconds.
+    #[error("retry after {0} seconds")]
+    RetryAfter(u16),
```
lann, line 59: This still seems to be arbitrarily splitting states between OK and error responses. "Retry after" seems like it should be an optional bit of data associated with the `unverified` (or `notFound`?) state.

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,124 @@
...
+pub enum CheckpointVerificationState {
+    /// The checkpoint signature is unverified and could be valid or invalid.
+    #[serde(rename_all = "camelCase")]
+    Unverified,
+    /// The checkpoint is verified.
+    #[serde(rename_all = "camelCase")]
+    Verified,
+    /// The checkpoint log length does not exist.
+    #[serde(rename_all = "camelCase")]
+    NotFound,
```
lann, line 29: I'm still unclear on the difference between `unverified` and `notFound`. I guess a concrete question to ask is: What would a client do differently if it received one vs the other?

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,124 @@
...
+        /// The error message
+        message: String,
+    },
+}
+
+impl MonitorError {
+    /// Returns the HTTP status code of the error.
+    pub fn status(&self) -> u16 {
+        match self {
+            Self::RetryAfter(_) => 503,
```
lann, line 74: 503 has strong "the server is broken" connotations.

calvinrp: `429`? Or just do inside of the JSON body of the `200` response?

crates/api/src/v1/monitor.rs
calvinrp: `unverified` is when the server is unable to verify whether valid or invalid. This may be terminal. Or, currently, there is a `503` that instructs to retry.

`notFound` is an invalid checkpoint due to the registry either never produced a checkpoint with the given `logLength` (between two other valid ones) or it is definitively in the future.

crates/api/src/v1/monitor.rs
lann: I think it makes more sense in the body.

crates/api/src/v1/monitor.rs
calvinrp: @esoterra was thinking the reason why something failed was important information.

Feels like we need to communicate the minimum options: `retry`, `verified`, `invalid`, `unknown`

lann: That's fine, its just that if *I* don't understand the difference it isn't likely that many other people will either. :slightly_smiling_face:

crates/api/src/v1/monitor.rs
calvinrp: Updated, from our discussion

crates/api/src/v1/monitor.rs
calvinrp: Updated, from our discussion

crates/api/src/v1/monitor.rs
calvinrp: Updated, from our discussion

crates/server/src/api/v1/mod.rs
```diff
@@ -90,6 +94,35 @@ pub async fn not_found() -> impl IntoResponse {
     }
 }
 
+/// An extractor for the `Warg-Registry` header.
```
esoterra, line 97: Can we extend this doc comment to mention that it currently returns an error if it's ever provided currently?

crates/server/src/api/v1/monitor.rs
```diff
@@ -0,0 +1,124 @@
...
+            CheckpointVerificationState::Unverified,
+        ),
+        Err(error) => return Err(MonitorApiError::from(error)),
+    };
+
+    // if `Unverified`, check signature against keys in operator log:
+    //
+    // - if `key_id` is not known or it does not have permission to sign checkpoints or
+    //   the `signature` is invalid, `signature` is `Invalid`;
+    if signature_verification == CheckpointVerificationState::Unverified {
```
esoterra, line 100: I'm not sure that this does what  you think it does, though I might be missing something.
Do you have any tests for this logic?

Specifically, I don't see when the `signature_verification` mut variable on line 60 is ever actually used.

crates/api/src/v1/monitor.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Represents a monitor API error.
+#[non_exhaustive]
+#[derive(Debug, Error)]
+pub enum MonitorError {
+    /// An error with a message occurred.
+    #[error("{message}")]
+    Message {
+        /// The HTTP status code.
```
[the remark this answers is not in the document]
calvinrp: Yes, existing pattern used throughout the API on the error messages. Pre-dates me. Open to refactoring in another PR.

crates/server/src/api/v1/mod.rs
calvinrp: Yes, added

crates/server/src/api/v1/monitor.rs
calvinrp: See lines 77-85 above in the file.

This logic is particular to the design of the reference implementation as it stands today. Every `TimestampedCheckpoint` that was issued is stored. We do a lookup and see if the requested checkpoint is an exact match first. That's where this `signature_verification` could be `Verified` for an exact match or `Unverified` if it is not an exact match. If it is not exact match, we proceed to see if the signature could be valid given what is in the operator log.

crates/protocol/tests/operator-logs/longer.json
```diff
@@ -37,4 +37,4 @@
             ]
         }
     }
-]
\ No newline at end of file
+]
```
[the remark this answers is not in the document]
calvinrp: I can add it back, if desired. But it probably would be better to not have the extra newline. Other than simplifying the diff.

crates/protocol/src/operator/state.rs
```diff
@@ -132,6 +156,17 @@ impl LogState {
         self.keys.get(key_id)
     }
 
+    /// Gets the namespace state. Expects the namespace to be lowercase.
+    pub fn namespace_state(&self, lowerspaced_namespace: &str) -> Option<&NamespaceState> {
```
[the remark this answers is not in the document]
calvinrp: It might be easier to talk thru, when you get a sec. I believe there is correct logic implementing, as is.

esoterra (approved):

PR #234, opened 2023-12-02
Make `download_content` public in `Client`.

peterhuene:
This PR makes `download_content` a public function from `Client`.

It turns out that this is a useful function to have for clients that can determine missing content hashes from the client storage and instruct the client to download just the content.

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -601,7 +601,11 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
         Ok(record)
     }
 
-    async fn download_content(&self, digest: &AnyHash) -> Result<PathBuf, ClientError> {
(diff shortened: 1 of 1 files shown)


crates/client/src/lib.rs
```diff
@@ -601,7 +601,11 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
         Ok(record)
     }
 
-    async fn download_content(&self, digest: &AnyHash) -> Result<PathBuf, ClientError> {
+    /// Downloads the content for the specified digest into client storage.
+    ///
```
calvinrp, line 605: `cargo fmt` is failing because of trailing space L605 in the doc comment

calvinrp (approved):
Approved, but might be good to fix the `cargo fmt` issue before merging.

PR #238, opened 2024-01-04
fix for client to skip the prove consistency request if checkpoint is unchanged

calvinrp:
Client will skip the prove consistency request if the previous checkpoint is still the latest checkpoint. This fixes an error when `from` and `to` are the same `log_length`.

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -539,16 +539,18 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
         }
 
         if let Some(from) = self.registry.load_checkpoint().await? {
-            self.api
-                .prove_log_consistency(
-                    ConsistencyRequest {
-                        from: from.as_ref().checkpoint.log_length,
-                        to: ts_checkpoint.as_ref().checkpoint.log_length,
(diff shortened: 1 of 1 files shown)


macovedj (approved):

crates/client/src/lib.rs
```diff
@@ -539,16 +539,18 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+                    .prove_log_consistency(
+                        ConsistencyRequest {
+                            from: from.as_ref().checkpoint.log_length,
+                            to: ts_checkpoint.as_ref().checkpoint.log_length,
+                        },
+                        Cow::Borrowed(&from.as_ref().checkpoint.log_root),
+                        Cow::Borrowed(&ts_checkpoint.as_ref().checkpoint.log_root),
+                    )
+                    .await?;
+            }
```
lann, line 553: Does this allow the registry to rewind the checkpoint?

calvinrp: That's a good point, I'll add another check and error for that in the client. Client should error if the registry checkpoint goes back in time.

PR #239, opened 2024-01-04
client will error if registry rewinds checkpoint log length

calvinrp:


crates/client/src/lib.rs
```diff
@@ -539,17 +540,29 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+                            ConsistencyRequest {
+                                from: from_log_length,
+                                to: to_log_length,
+                            },
+                            Cow::Borrowed(&from.as_ref().checkpoint.log_root),
+                            Cow::Borrowed(&ts_checkpoint.as_ref().checkpoint.log_root),
+                        )
+                        .await?
+                }
+                Ordering::Equal => {}
```
lann, line 565: Does this need to check for total checkpoint equality? (I'm really not sure; just want to be sure that the registry can't swap in another checkpoint with the same length)

calvinrp: Possible it would be caught during other parts of the inclusion proofs. But better to make it an explicit check and error. Pushed an update.

lann (approved):

PR #243, opened 2024-02-07
Fix transaction implementation in `LogState::validate`.

### LINKED ISSUE #242: LogState methods `snapshot` and `rollback` are not implemented correctly
Currently the `LogState::snapshot` and `LogState::rollback` methods are implemented as simply recording the length of the various data structures on snapshot and truncating back to the original lengths for rollback; the intention of this was to prevent a deep clone of the entire state, which may be expensive.

While this works when the log entries are purely additive, if the log entry is a _revoke permission_ or, in the case of a package log state, a _yank release_, then this scheme does not restore the original log state; in these cases, interior data is modified and the lengths of the log state data structures do not change.

I think the best fix for this would be:

1) Make `validate` take `self` and return `Result<Self, ValidationError>` and remove the incorrect `snapshot` and `rollback` implementations.
2) For the in-memory data store used for testing, clone the log state before validation, so that on error, the persistent `LogState` is unmodified.
3) For the postgres data store, a clone isn't required as the in-memory log state is ephemeral.

peterhuene:
This PR removes the incorrect transaction implementation for `LogState::validate`.

Now `validate` takes ownership of `self` and returns `Result<Self, ValidationError>`.

This means that callers that expect to keep the log state following an invalid log entry must clone the state prior to validation.

As the in-memory data store is the only store that persists the log state in memory, it now clones the state before validation and updates the log state upon successful validation.

For the postgres data store, the log state was loaded from the database and is discarded on error, so no clone is necessary.

Fixes #242.

### DIFF
--- a/crates/server/src/datastore/postgres/mod.rs
+++ b/crates/server/src/datastore/postgres/mod.rs
@@ -224,7 +224,7 @@ where
     conn.transaction::<_, DataStoreError, _>(|conn| {
         async move {
             // Get the record content and validator
-            let (id, content, mut validator) = schema::records::table
+            let (id, content, validator) = schema::records::table
                 .inner_join(schema::logs::table)
                 .select((
                     schema::records::id,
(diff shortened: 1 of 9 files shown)


crates/server/src/datastore/postgres/mod.rs
```diff
@@ -251,12 +251,12 @@ where
             })?;
 
             // Validate the record
-            validator.validate(&record).map_err(Into::into)?;
+            let validator = validator.0.validate(&record).map_err(Into::into)?;
```
peterhuene, line 254: Note that I didn't bother renaming "validator" here as the column name is still "validator" and not "state"; I figured it wasn't worth a migration.

calvinrp (approved):

PR #247, opened 2024-02-16
Namespaces

macovedj:
This PR adds namespace config to warg client for federation.
The client will check operator logs for imported namespaces and use them as warg headers on requests.
The client will check for warg hint headers when fetching logs, and prompt users to update local namespace config to point at a domain if the server provides a hint.
If the client does not find a package namespace in operator logs, it will check the local namespace config to see if there is a registry domain that it should include as a warg header.


crates/client/src/lib.rs
```diff
@@ -533,11 +619,59 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
                     packages: Cow::Borrowed(&last_known),
                 })
                 .await
-                .map_err(|e| {
-                    ClientError::translate_log_not_found(e, |id| {
+            {
+                Ok(res) => res,
+                Err(api::ClientError::LogNotFoundWithHint(log_id, header)) => {
```
lann, line 624: This kind of user interaction logic probably doesn't belong in this client library, at least not in this low-level method. The simplest place to put it would be in the warg bin code or - if it needs to be reused by other dependants - at least in a separate method or gated behind an opt-in flag of some sort.

crates/client/src/lib.rs
```diff
@@ -533,11 +619,59 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+                    let namespace = terms.next();
+                    let registry = terms.next();
+                    let resp = if let (Some(namespace), Some(registry)) = (namespace, registry) {
+                        print!(
+                                      "One of the packages you're requesting does not exist in the registry you're using.
+                      However, the package namespace `{namespace}` does exist in the registry at {registry}.\nWould you like to configure your warg cli to use this registry for packages with this namespace in the future? y/N\n",
+                                  );
+                        std::io::Write::flush(&mut std::io::stdout()).expect("flush failed!");
+                        let mut buf = String::new();
+                        std::io::stdin().read_line(&mut buf).unwrap();
```
lann, line 636: You really don't want to use `read_line` for interactive user input. We already have a dependency on `dialoguer` for this sort of thing; see https://github.com/bytecodealliance/registry/blob/main/src/commands/key.rs#L181

lann (commented):
Partial review

crates/client/src/storage.rs
```diff
@@ -32,33 +33,57 @@ pub trait RegistryStorage: Send + Sync {
     /// Reset registry local data
     async fn reset(&self, all_registries: bool) -> Result<()>;
 
+    // /// Directory where all registries are stored
+    // fn registries_dir(&self) -> PathBuf;
     /// Loads most recent checkpoint
-    async fn load_checkpoint(&self) -> Result<Option<SerdeEnvelope<TimestampedCheckpoint>>>;
+    async fn load_checkpoint(
+        &self,
+        namespace_registry: &Option<HeaderValue>,
```
lann, line 41: I'd highly recommend wrapping `HeaderValue` in a new type like `RegistryName` (or `RegistryDomain`?); you really don't want your trait interfaces to depend on a specific external library type like this. Something like:

```rust
pub struct RegistryName(String);

impl TryFrom<HeaderValue> for RegistryName ...

impl From<RegistryName> for HeaderValue ...
```

crates/client/src/storage.rs
```diff
@@ -108,6 +133,20 @@ pub trait ContentStorage: Send + Sync {
...
+/// Namespace Map storage data must be synchronized if shared between
+/// multiple threads an
+#[async_trait]
+pub trait NamespaceMapStorage: Send + Sync {
+    /// Loads namespace map
+    async fn load_namespace_map(&self) -> Result<Option<HashMap<String, String>>>;
+    /// Reset namespace mappings
+    async fn reset_namespaces(&self) -> Result<()>;
+    /// Store namespace mapping
+    async fn store_namespace(&self, namespace: String, registry_domain: String) -> Result<()>;
```
lann, line 147: The fact that `registry_domain` is a `String` here rather than a `HeaderValue` is another strong hint that it should have its own type.

crates/client/src/storage/fs.rs
```diff
@@ -42,10 +47,13 @@ impl FileSystemRegistryStorage {
     /// If the lock cannot be acquired, `Ok(None)` is returned.
     pub fn try_lock(base_dir: impl Into<PathBuf>) -> Result<Option<Self>> {
         let base_dir = base_dir.into();
+        let mut registries_dir = base_dir.clone();
+        registries_dir.pop();
```
lann, line 51: This can be silently incorrect if `base_dir` is empty.

```suggestion
        let registries_dir = base_dir.parent()
            .context("base_dir cannot be empty")?
            .to_path_buf();
        registries_dir.pop();
```

crates/client/src/storage/fs.rs
```diff
@@ -60,17 +68,41 @@ impl FileSystemRegistryStorage {
     pub fn lock(base_dir: impl Into<PathBuf>) -> Result<Self> {
         let base_dir = base_dir.into();
         let lock = FileLock::open_rw(base_dir.join(LOCK_FILE_NAME))?;
+        let mut registries_dir = base_dir.clone();
+        registries_dir.pop();
```
lann, line 72: ditto above

crates/client/src/storage/fs.rs
```diff
@@ -328,6 +435,49 @@ impl ContentStorage for FileSystemContentStorage {
...
+        if let Some(mut mapping) = mapping {
+            mapping.insert(namespace, registry_domain);
+            let json = serde_json::to_string(&mapping)?;
+            fs::write(&self.base_dir, json)?;
+        } else {
+            let mut mapping = HashMap::new();
+            mapping.insert(namespace, registry_domain);
+            let json = serde_json::to_string(&mapping)?;
+            fs::write(&self.base_dir, json)?;
+        }
```
lann, line 476: I think this should work:

```suggestion
        let mut mapping = self.load_namespace_map()
            .await?
            .unwrap_or_default();
        mapping.insert(namespace, registry_domain);
        let json = serde_json::to_string(&mapping)?;
        fs::write(&self.base_dir, json)?;
```

src/commands/info.rs
```diff
@@ -25,21 +32,28 @@ impl InfoCommand {
 
         println!("registry: {url}", url = client.url());
         println!("\npackages in client storage:");
```
lann, line 34: Isn't this header incorrect with --namespaces? Maybe the whole thing below should be inverted:

```suggestion
        if self.namespaces {
            Self::print_namespace_map(&client).await?;
            return Ok(());
        }
        println!("\npackages in client storage:");
```

src/commands/publish.rs
```diff
@@ -116,17 +116,25 @@ pub struct PublishInitCommand {
...
+        if let Some(retry) = retry {
+            retry.store_namespace(&client).await?
+        }
+        client.refresh_namespace(self.name.namespace()).await?;
 
+        let signing_key = if let Some(nm) = client.get_warg_header() {
+            self.common.signing_key(&RegistryUrl::new(nm.to_str()?)?)?
+        } else {
+            self.common.signing_key(client.url())?
+        };
```
lann, line 131: It looks like this snippets are repeated many times below. Can these two bits of logic be moved into `CommonOptions`? Would require some signature changes (or new methods) like:

```rust
pub async fn create_client(&self, config: &Config, retry: Option<Retry>) ...

pub fn signing_key(&self, client: &Client) -> Result<PrivateKey> {
```

crates/client/src/lib.rs
```diff
@@ -692,20 +783,56 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+    }
+
+    async fn update_checkpoints<'a>(
+        &mut self,
+        ts_checkpoints: HashMap<std::string::String, SerdeEnvelope<TimestampedCheckpoint>>,
+        mut packages: HashMap<String, Vec<PackageInfo>>,
+    ) -> Result<(), ClientError> {
+        for (name, ts_checkpoint) in ts_checkpoints {
+            if self.url().safe_label() != name {
+                self.map_warg_header(&Some(HeaderValue::from_str(&name).unwrap()));
```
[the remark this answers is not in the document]
macovedj: Yeah I also didn't love this, but keeping client state helps avoid passing namespace registry values through all the functions everywhere.  This should be the only place where it's mutated.  Also the client state is reset based on local storage for each command, so in the error scenario here, it should just panic, and you'll have a fresh client state next time you run the `update` command.  If we really hate it, this is also actually new functionality, the client doesn't currently offer the ability to update all packages across all registries found in client storage.  So we could also just require users to be explicit about which registry they want to update packages for.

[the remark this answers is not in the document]
macovedj: I guess when I say only place where it's mutated, other than the very first `refresh_namespace` call at the beginning of each command

crates/client/src/lib.rs
```diff
@@ -97,6 +124,49 @@ impl<R: RegistryStorage, C: ContentStorage> Client<R, C> {
...
+                } else {
+                    self.api.map_warg_header(None);
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Get namespace registry
+    pub fn get_warg_header(&self) -> &Option<HeaderValue> {
```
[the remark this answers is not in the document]
macovedj: I think this one does.  A few commands use it to load package info and get signing keys.

crates/client/src/storage/fs.rs
```diff
@@ -47,13 +47,16 @@ impl FileSystemRegistryStorage {
     /// If the lock cannot be acquired, `Ok(None)` is returned.
     pub fn try_lock(base_dir: impl Into<PathBuf>) -> Result<Option<Self>> {
         let base_dir = base_dir.into();
-        let mut registries_dir = base_dir.clone();
+        let registries_dir = &mut base_dir
+            .parent()
+            .context("base_dir cannot be empty")?
+            .to_path_buf();
         registries_dir.pop();
```
lann, line 54: Sorry, my suggestion was off; this `pop` needs to be removed too; `.parent()` does the same thing.

lann (approved):

PR #248, opened 2024-02-20
Auth token

macovedj:
This PR adds login and logout commands to add or delete auth tokens to your keyring and then adds them as bearer tokens to requests made to your warg server


crates/client/src/api.rs
```diff
@@ -172,25 +171,46 @@ impl WithWargHeader for RequestBuilder {
...
 impl Client {
     /// Creates a new API client with the given URL.
-    pub fn new(url: impl IntoUrl) -> Result<Self> {
+    pub fn new(url: impl IntoUrl, auth_token: Option<Secret<String>>) -> Result<Self> {
         let url = RegistryUrl::new(url)?;
         Ok(Self {
             url,
             client: reqwest::Client::new(),
             warg_registry_header: None,
+            auth_token,
```
lann, line 205: I'd recommend using https://docs.rs/reqwest/0.11.24/reqwest/struct.ClientBuilder.html#method.default_headers to replace all the changes in this file. The first example in the method docs there is basically what you would need (plus prepending `Bearer ` to the value).

src/commands.rs
```diff
@@ -81,16 +90,23 @@ impl CommonOptions {
         config: &Config,
         retry: Option<Retry>,
     ) -> Result<FileSystemClient, ClientError> {
-        let client = match FileSystemClient::try_new_with_config(self.registry.as_deref(), config)?
-        {
+        let client = match FileSystemClient::try_new_with_config(
+            self.registry.as_deref(),
+            config,
+            self.auth_token(config)?.map(Secret::from),
```
lann, line 96: Looks like `auth_token` already returns a `Secret`? :thinking: 

```suggestion
            self.auth_token(config)?,
```

src/commands/login.rs
```diff
@@ -0,0 +1,50 @@
...
+        } else {
+            bail!("Please configure your home registry: warg config --registry <registry-url>")
+        }
+    }
+}
+
+impl LoginCommand {
+    /// Executes the command.
+    pub async fn exec(self) -> Result<()> {
+        let token = rpassword::prompt_password("Enter auth token:\n")
```
lann, line 43: Not sure why we didn't use https://docs.rs/dialoguer/0.11.0/dialoguer/struct.Password.html before... :thinking:

I guess for consistency we should use either rpassword or dialoguer in both places, so up to you if you want to change both in this PR.

crates/client/src/api.rs
macovedj: Think I agree with Calvin here... though if there's something in the middle, the default headers is definitely a smaller change.  Went ahead and pushed the other suggestions.

lann (approved):

Issue #221, opened 2023-10-20
`accept_ranges` field is not camelCased in JSON


calvinrp:
Serde `rename_all` as an attribute on an enum is renaming all the variants of the enum, not all the fields of those variants.

[Missing `rename_all` on enum variant](https://github.com/bytecodealliance/registry/blob/release-0.2/crates/api/src/v1/mod.rs#L16)

lann:
We're going to use this as an exercise in change management for `0.2.x` releases.

As a general approach, I'm proposing:

1. a serde `alias` for the corrected name; this gives clients forward-compatibility
2. wait a release
3. change the field default to the correct name, alias the incorrect name
4. wait (a couple releases? until `0.3`?)
5. drop the alias

calvinrp:
Merged in #253

PR #253, opened 2024-02-29
removed temporary aliases in struct fields with serde

calvinrp:
Drop temporary aliasing, per release policy  #221

### DIFF
--- a/crates/api/src/v1/mod.rs
+++ b/crates/api/src/v1/mod.rs
@@ -28,13 +28,7 @@ pub enum ContentSource {
         /// The URL of the content.
         url: String,
         /// Optional, server accepts for HTTP Range header.
-        /// TODO remove rename, see issue: https://github.com/bytecodealliance/registry/issues/221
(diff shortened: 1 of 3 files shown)


crates/api/src/v1/mod.rs
```diff
@@ -28,13 +28,7 @@ pub enum ContentSource {
         /// The URL of the content.
         url: String,
         /// Optional, server accepts for HTTP Range header.
-        /// TODO remove rename, see issue: https://github.com/bytecodealliance/registry/issues/221
-        #[serde(
-            default,
-            skip_serializing_if = "is_false",
-            rename = "accept_ranges",
-            alias = "acceptRanges"
```
lann, line 36: This one is technically breaking "forward compatibility" since 0.3.0 code is still serializing as `accept_ranges`. Better to just swap the `rename` and `alias` values now and drop the `alias` later.

lann (approved):

PR #257, opened 2024-03-06
Warg credentials

macovedj:
Adds a new crate exposing methods for accessing credentials in the keyring so that consumers don't need to implement it too.

This also checks for auth tokens less eagerly, avoiding a keyring password if the user hasn't set an auth token.

### DIFF
--- a/crates/client/src/config.rs
+++ b/crates/client/src/config.rs
@@ -108,12 +108,12 @@ pub struct Config {
     pub namespace_map_path: Option<PathBuf>,
 
     /// List of creds availabe in keyring
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub keys: Option<IndexSet<String>>,
+    #[serde(default, skip_serializing_if = "IndexSet::is_empty")]
+    pub keys: IndexSet<String>,
 
     /// Whether or not an auth key should be retreived from keyring
     #[serde(default)]
-    pub auth: bool,
+    pub keyring_auth: bool,
 }
 
 impl Config {
@@ -191,7 +191,7 @@ impl Config {
                 pathdiff::diff_paths(&p, &parent).unwrap()
             }),
             keys: self.keys.clone(),
-            auth: false,
+            keyring_auth: false,
         };
 
         serde_json::to_writer_pretty(
--- a/crates/client/src/config.rs
+++ b/crates/client/src/config.rs
@@ -110,6 +110,10 @@ pub struct Config {
     /// List of creds availabe in keyring
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub keys: Option<IndexSet<String>>,
+
+    /// Whether or not an auth key should be retreived from keyring
+    #[serde(default)]
+    pub auth: bool,
 }
 
 impl Config {
@@ -187,6 +191,7 @@ impl Config {
                 pathdiff::diff_paths(&p, &parent).unwrap()
             }),
             keys: self.keys.clone(),
+            auth: false,
         };
 
         serde_json::to_writer_pretty(
--- a/crates/client/src/config.rs
+++ b/crates/client/src/config.rs
@@ -191,7 +191,7 @@ impl Config {
                 pathdiff::diff_paths(&p, &parent).unwrap()
             }),
             keys: self.keys.clone(),
-            keyring_auth: false,
+            keyring_auth: self.keyring_auth,
         };
 
         serde_json::to_writer_pretty(
--- a/crates/credentials/Cargo.toml
+++ b/crates/credentials/Cargo.toml
@@ -13,7 +13,7 @@ repository = { workspace = true}
 [dependencies]
 anyhow = { workspace = true}
 indexmap = { workspace = true}
-keyring = "2.3.2"
+keyring = { workspace = true }
 secrecy = { workspace = true}
 warg-client = { workspace = true}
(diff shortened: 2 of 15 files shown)


crates/client/src/config.rs
```diff
@@ -110,6 +110,10 @@ pub struct Config {
     /// List of creds availabe in keyring
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub keys: Option<IndexSet<String>>,
+
+    /// Whether or not an auth key should be retreived from keyring
+    #[serde(default)]
+    pub auth: bool,
```
lann, line 116: Maybe `keyring_auth`?

crates/credentials/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+license = { workspace = true}
+homepage = { workspace = true}
+repository = { workspace = true}
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = { workspace = true}
+indexmap = { workspace = true}
+keyring = "2.3.2"
```
peterhuene, line 16: Nit: although this is probably the only crate that would take this dependency, I've generally favored having dependency versions kept at the workspace-level.

crates/client/src/config.rs
```diff
@@ -110,6 +110,10 @@ pub struct Config {
     /// List of creds availabe in keyring
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub keys: Option<IndexSet<String>>,
```
peterhuene, line 112: While this is preexisting, this should be `IndexSet<String>` with `skip_serializing_if` set to `IndexSet::is_empty`.

Having it wrapped in an option is a bit unwieldy.

src/commands.rs
```diff
@@ -117,20 +115,23 @@ impl CommonOptions {
         };
         let config = self.read_config()?;
         get_signing_key(
-            &registry_url.map(|reg| reg.safe_label()),
-            config.keys.expect("Please set a default signing key by typing `warg key set <alg:base64>` or `warg key new"),
-            config.home_url,
+            registry_url.map(|reg| reg.safe_label()).as_deref(),
+            &config.keys.expect("Please set a default signing key by typing `warg key set <alg:base64>` or `warg key new"),
```
peterhuene, line 119: Preexisting issue:

```suggestion
            &config.keys.context("please set a default signing key by typing `warg key set <alg:base64>` or `warg key new")?,
```

Ideally just `&config.keys` if we remove the wrapping `Option`.

src/commands/key.rs
```diff
@@ -159,8 +162,7 @@ impl KeyDeleteCommand {
             .with_prompt("are you sure you want to delete your signing key")
             .interact()?
         {
-            // delete_signing_key(&self.common.registry, config)?;
-            delete_signing_key(&self.common.registry, config.keys.clone().expect("Please set a default signing key by typing `warg key set <alg:base64>` or `warg key new"), config.home_url.clone())?;
+            delete_signing_key(self.common.registry.as_deref(), config.keys.as_ref().expect("Please set a default signing key by typing `warg key set <alg:base64>` or `warg key new"), config.home_url.as_deref())?;
```
peterhuene, line 165: Same comment as above (shouldn't use `expect` here):

```suggestion
            delete_signing_key(self.common.registry.as_deref(), config.keys.as_ref().context("please set a default signing key by typing `warg key set <alg:base64>` or `warg key new")?, config.home_url.as_deref())?;
```

src/commands/logout.rs
```diff
@@ -41,6 +41,8 @@ impl LogoutCommand {
     pub async fn exec(self) -> Result<()> {
         self.keyring_entry
             .delete_entry(self.common.read_config()?.home_url)?;
+        let mut config = self.common.read_config()?;
+        config.auth = false;
```
peterhuene, line 45: What do these lines accomplish?

peterhuene (commented):
Looks great overall (and really happy to be reusing `warg-credentials` from `cargo-component` soon).

Just a few feedback comments, mostly nit and mostly preexisting issues.

src/commands/logout.rs
macovedj: So a lot of the recent work in the warg cli is dealing with the fact that in order to figure out whether a credential in your keyring even exists, there's a password prompt.

Basically this bool, along with the list of signing key names that was added in prior prs is stored in the config, and later the warg client is able to determine via config that keyring access shouldn't be attempted, so that it can be more intentional about keyring access when we know that it will fail.

Here specifically logging out removes your auth token from your keyring, so the config is updated for the reason described above.  I do realize that I'm missing a `config.write_to_file` here, so I can add that along with the other recommended changes.

peterhuene: I figured we were failing to save the config back out; thanks for the confirmation!

crates/client/src/config.rs
macovedj: Oh thanks for the rec I think this will actually make a lot of the keyring code a lot simpler.

peterhuene (approved):
LGTM!

peterhuene:
I just noticed that we should add the new crate to https://github.com/bytecodealliance/registry/blob/main/ci/publish.rs#L16 and https://github.com/bytecodealliance/registry/blob/main/ci/publish.rs#L31.

Issue #261, opened 2024-04-03
Provide optional warning messages on the `/v1/fetch/logs` response


calvinrp:
When the registry wants to provide warnings to the client CLI, such as for deprecated packages, we need to add an optional field in the fetch logs response.

[See the code for FetchLogsResponse](https://github.com/bytecodealliance/registry/blob/main/crates/api/src/v1/fetch.rs#L42-L55)
```rust
pub struct FetchLogsResponse {
    /// Whether there are more records to fetch.
    #[serde(default)]
    pub more: bool,
    /// The operator records appended since the last known operator record.
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub operator: Vec<PublishedRecord>,
    /// The package records appended since last known package record ids.
    #[serde(default, skip_serializing_if = "IndexMap::is_empty")]
    pub packages: IndexMap<LogId, Vec<PublishedRecord>>,
}
```

Could add a field
```rust
    /// Optional warning message.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub warning: Option<String>,
```

Then in the client CLI, print the warning message, if present.
[See](https://github.com/bytecodealliance/registry/blob/main/crates/client/src/lib.rs#L620-L636)

lann:
I would give this a little more structure as future-proofing, e.g.
```json
"warnings": [{"message": "..."}]
```

RobbieMcKinstry:
LGTM! I'm happy to tackle this one. I'll put something together in a few hours. 👍🏻

PR #262, opened 2024-04-05
Add warning field to fetch logs response.

RobbieMcKinstry:
The intention of adding this field is to handle moved repositories. When a repository is moved, Warg will use the new repository name instead, but the users must be alerted so they can update their dependency list.

Closes https://github.com/bytecodealliance/registry/issues/261

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -629,6 +629,11 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
                     packages: Cow::Borrowed(&last_known),
                 })
                 .await
+                .inspect(|res| {
+                    for warning in res.warnings.iter() {
+                        tracing::warn!("WARNING: {}", warning.message);
+                    }
+                })
                 .map_err(|e| {
                     ClientError::translate_log_not_found(e, |id| {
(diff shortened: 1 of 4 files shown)


RobbieMcKinstry:
cc: @calvinrp

crates/client/src/lib.rs
```diff
@@ -629,6 +629,11 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
                     packages: Cow::Borrowed(&last_known),
                 })
                 .await
+                .inspect(|res| {
+                    for warning in res.warnings.iter() {
+                        tracing::warn!("WARNING: {}", warning.message);
```
lann, line 634: Nit: `WARNING:` here is probably unnecessary; `tracing` subscribers will typically annotate with the event level already.

It would probably be more helpful to describe the _type_ of warning, e.g.
```suggestion
                        tracing::warn!("Fetch warning from registry: {}", warning.message);
```

RobbieMcKinstry: Yeah, that tracks as far as I'm concerned. 👍🏻

lann (approved):

lann:
Thanks!

PR #263, opened 2024-04-15
make better use of namespace storage

macovedj:
Rather than updating the state at the beginning of a client interaction to use a specific url, the client expects to check the namespace storage for a url and pass it as an arg, better enabling consumers to fetch packages across multiple namespaces.

There is also updated logic for encouraging standard ways of retrying with `Warg-Hint-Header` responses from the registry after using the hint to update local namespace mappings.

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -10,7 +10,6 @@ use secrecy::Secret;
 use semver::{Version, VersionReq};
 use std::cmp::Ordering;
 use std::fs;
-use std::future::Future;
 use std::str::FromStr;
 use std::{borrow::Cow, path::PathBuf, time::Duration};
 use storage::{
@@ -121,13 +120,7 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
                 Ok(Some(warg_protocol::operator::NamespaceState::Defined)) => {
                     return Ok(None);
                 }
-                Ok(None) => return Ok(None),
-                Err(e) => {
-                    return Err(ClientError::SimilarNamespace {
-                        namespace: namespace.to_string(),
-                        e: e.to_string(),
-                    });
-                }
+                _ => (),
             }
         };
         let nm_map = self.namespace_map.load_namespace_map().await?;
@@ -1285,13 +1278,4 @@ impl Retry {
             .await?;
         Ok(())
     }
-}
-
-/// Interactively retry when hint header received from warg server
-pub async fn with_interactive_retry<F>(f: impl Fn(Option<Retry>) -> F) -> Result<()>
-where
-    F: Future<Output = Result<()>>,
-{
-    f(None).await?;
-    Ok(())
 }
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -10,6 +10,7 @@ use secrecy::Secret;
 use semver::{Version, VersionReq};
 use std::cmp::Ordering;
 use std::fs;
+use std::future::Future;
 use std::str::FromStr;
 use std::{borrow::Cow, path::PathBuf, time::Duration};
 use storage::{
@@ -65,7 +66,7 @@ where
 impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C, N> {
     /// Creates a new client for the given URL, registry storage, and
     /// content storage.
-    pub fn new(
+    pub async fn new(
         url: impl IntoUrl,
         registry: R,
         content: C,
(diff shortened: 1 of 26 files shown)


crates/client/src/lib.rs
```diff
@@ -101,6 +102,36 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+            let namespace_state = op.state.namespace_state(namespace);
+            if let Ok(Some(warg_protocol::operator::NamespaceState::Imported { registry })) =
+                namespace_state
+            {
+                return Ok(Some(RegistryDomain::from_str(registry)?));
+            } else if let Ok(Some(warg_protocol::operator::NamespaceState::Defined)) =
+                namespace_state
+            {
+                return Ok(None);
+            };
```
lann, line 123: nit: You'd typically use `match` here:

```suggestion
            match op.state.namespace_state(namespace) {
                Ok(Some(warg_protocol::operator::NamespaceState::Imported { registry })) => {
                    return Ok(Some(RegistryDomain::from_str(registry)?));
                }
                Ok(Some(warg_protocol::operator::NamespaceState::Defined)) => {
                    return Ok(None);
                }
                _ => (),
            }
```

crates/client/src/lib.rs
```diff
@@ -101,6 +102,36 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+    /// Get
+    pub async fn get_warg_registry(&self, namespace: &str) -> Result<Option<RegistryDomain>> {
+        self.update_checkpoint(None, &self.api.latest_checkpoint(None).await?, vec![])
+            .await?;
+        let operator = self
+            .registry()
+            .load_operator(Some(&RegistryDomain::from_str(namespace)?))
+            .await?;
+        if let Some(op) = operator {
+            let namespace_state = op.state.namespace_state(namespace);
```
lann, line 114: Are `Err`s expected here? Should probably at least log them if not propagating, unless there is a specific reason to completely suppress.

crates/client/src/lib.rs
```diff
@@ -485,18 +492,64 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
-                    .await?
-                    .unwrap_or_else(|| PackageInfo::new(package.clone())),
-            );
+            let namespace = package.namespace();
+            let namespace_packages = namespaced.get_mut(namespace);
+            if let Some(nm_pkgs) = namespace_packages {
+                nm_pkgs.push(package);
+            } else {
+                namespaced.insert(namespace, vec![package]);
+            }
```
lann, line 503: nit: There is a nice pattern for this:

```suggestion
            namespaced.entry(namespace).or_default().push(package);
```

src/bin/warg.rs
```diff
@@ -46,27 +46,98 @@ async fn main() -> Result<()> {
...
-        if let Some(e) = e.downcast_ref::<ClientError>() {
-            describe_client_error_or_retry(e).await?;
-        } else {
-            eprintln!("error: {e:?}");
+    match &WargCli::parse() {
+        WargCli::Config(cmd) => cmd.clone().exec().await?,
+        WargCli::Info(cmd) => cmd.clone().exec().await?,
+        WargCli::Key(cmd) => cmd.clone().exec().await?,
+        WargCli::Lock(cmd) => {
+            with_interactive_retry(|retry: Option<Retry>| async {
```
lann, line 54: Can this go outside the `match`? The commands that don't support `retry` shouldn't be returning this error anyway and can just ignore the `retry` variable?

lann (commented):
Only maybe-blocker is question about error handling.

crates/client/src/lib.rs
```diff
@@ -101,6 +102,44 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+                    return Ok(None);
+                }
+                Ok(None) => return Ok(None),
+                Err(e) => {
+                    return Err(ClientError::SimilarNamespace {
+                        namespace: namespace.to_string(),
+                        e: e.to_string(),
+                    });
+                }
+            }
```
lann, line 131: minor: its difficult to see that this `if` block always `return`s
```suggestion
            return match op.state.namespace_state(namespace) {
                Ok(Some(warg_protocol::operator::NamespaceState::Imported { registry })) => {
                    Ok(Some(RegistryDomain::from_str(registry)?))
                }
                Ok(Some(warg_protocol::operator::NamespaceState::Defined)) => {
                    Ok(None)
                }
                Ok(None) => return Ok(None),
                Err(e) => {
                    Err(ClientError::SimilarNamespace {
                        namespace: namespace.to_string(),
                        e: e.to_string(),
                    })
                }
            }
```

crates/client/src/lib.rs
```diff
@@ -995,6 +1059,15 @@ pub struct PackageDownload {
 /// Represents an error returned by Warg registry clients.
 #[derive(Debug, Error)]
 pub enum ClientError {
+    /// Similar Namespace
+    #[error("Namespace `{namespace}` not found in operator log but found namespace `{e}`, which has alternative casing.")]
```
lann, line 1063: Are we not enforcing lowercase namespaces? I think we should be... https://github.com/WebAssembly/component-model/pull/338

crates/client/src/lib.rs
```diff
@@ -1182,3 +1255,43 @@ impl ClientError {
...
+}
+
+/// Interactively retry when hint header received from warg server
+pub async fn with_interactive_retry<F>(f: impl Fn(Option<Retry>) -> F) -> Result<()>
+where
+    F: Future<Output = Result<()>>,
+{
+    f(None).await?;
+    Ok(())
+}
```
lann, line 1297: This doesn't appear to do what it claims to do?

crates/client/src/lib.rs
calvinrp: Yeah, that is outstanding. Probably better as a separate PR.

calvinrp:
Collaborated with @macovedj to rework these changes into PR #268 which was merged. Thus closing this PR.

PR #269, opened 2024-05-05
Improved package init and publish flow

calvinrp:
When publishing either a package `Release` or `Init`, a new error `ConflictPendingPublish(RecordId)` may be returned.

This instructs the client that it could wait for the conflicting pending record (on the same `LogId` package) to complete and then retry to publish.

If a registry service offers key management and needs to sign the `Init` record, this enables a smoother experience and messaging on the CLI. Where the registry would return the `RecordId` of the registry signed `Init` record as a `ConflictPendingPublish` error, the CLI would wait for that record to publish and then retry. Since the package is already initialized, an error message would be presented that indicates that the package was initialized but not with the `Init` record that you signed.

Also, added a feature flag for `cli-interactive` for the `warg-client` and a config flag to disable dialoguer interactive prompting even when that feature is enabled.


tests/client.rs
```diff
@@ -18,7 +18,7 @@ fn create_client(config: &Config) -> Result<FileSystemClient> {
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn client_incrementally_fetches() -> Result<()> {
-    const RELEASE_COUNT: usize = 300;
+    const RELEASE_COUNT: usize = 10;
```
macovedj, line 21: Why is this release count reduced in this test?

calvinrp: The test takes too long to run and times out with the changes. Also, kind of unrealistic test at `300` releases.

crates/client/src/lib.rs
```diff
@@ -310,109 +312,161 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+                            format!(
+                                "Package `{package_name}` was not found.
+If it exists, you may not have access.
+Attempt to create `{package_name}` and publish the release y/N\n",
+                                package_name = &info.name,
+                            )
+                        } else {
+                            format!(
+                                "Package `{package_name}` was not found.
+If it exists, you may not have access without logging in. Try: `warg login`
```
macovedj, line 370: Should we suggest to users that they log in and immediately prompt them to publish without making them log in?  Also guessing that if the registry they're using requires auth, that that'll show up in the error?  Not sure what best DX is here, but might want to consider just using the same prompt for both cases

Cargo.toml
```diff
@@ -58,6 +58,7 @@ testresult = "0.3.0"
 [features]
 default = []
 postgres = ["warg-server/postgres"]
+not-cli = ["warg-client/not-cli"]
```
lann, line 61: Negative features are frowned upon; if a crate has multiple (transitive) paths to a dependency (e.g. a direct dependency plus a transitive dependency), the resolved feature set will always be the intersection of all required features, which is likely to cause breakage with negative features.

I'd recommend just making this `cli` (`cli-interactive`?) and adding it to the `default` set.

crates/api/src/v1/package.rs
```diff
@@ -301,6 +310,16 @@ impl<'de> Deserialize<'de> for PackageError {
             },
             RawError::Conflict { status: _, ty, id } => match ty {
                 EntityType::NamespaceImport => Ok(Self::NamespaceImported(id.into_owned())),
+                EntityType::Record => Ok(Self::ConflictPendingPublish(
+                    id.parse::<AnyHash>()
```
lann, line 314: I personally find this easier to read but :shrug: 
```suggestion
                    AnyHash::from_str(&id)
```

crates/client/src/config.rs
```diff
@@ -122,6 +122,10 @@ pub struct Config {
     /// Auto accept registry hint or ask the user to confirm
     #[serde(default)]
     pub auto_accept_federation_hints: bool,
+
+    /// Disable dialoguer prompts.
+    #[serde(default)]
+    pub disable_dialoguer: bool,
```
lann, line 128: I wouldn't reference the specific library here. Maybe `disable_interaction`?

lann (commented):
Negative feature comment is only blocker

Cargo.toml
calvinrp: Thanks, will fix

Cargo.toml
```diff
@@ -56,9 +54,10 @@ wit-parser = "0.13.1"
 testresult = "0.3.0"
 
 [features]
-default = []
+default = ["cli-interactive"]
 postgres = ["warg-server/postgres"]
-not-cli = ["warg-client/not-cli"]
+cli-interactive = ["warg-client/cli-interactive"]
+not-interactive-cli = []
```
lann, line 60: Unused?

Cargo.toml
calvinrp: Fixed. Changed to `cli-interactive` feature flag, added it to `default` features.

But also added `not-cli-interactive` feature that doesn't include that feature `cli-interactive` since `default` does. Is that the correct way to set that option?

crates/api/src/v1/package.rs
calvinrp: Fixed. Revised other occurrences as well.

crates/client/src/config.rs
calvinrp: Fixed

crates/client/Cargo.toml
```diff
@@ -10,8 +10,10 @@ homepage = { workspace = true }
 repository = { workspace = true}
 
 [features]
-default = []
+default = ["cli-interactive"]
 native-tls-vendored = ["reqwest/native-tls-vendored"]
+cli-interactive = ["dialoguer"]
```
lann, line 15: nit; not much functional difference but prevents an implicit feature named "dialoguer" from being created

```suggestion
cli-interactive = ["dep:dialoguer"]
```

Cargo.toml
calvinrp: Is there a good way to specify that you don't want the `cli-interactive` feature if there isn't an empty set?

Cargo.toml
lann: When you want to disable any default features you can use `default-features = false` in the dependency spec, sometimes in combination with default features you want to retain, e.g.:

```toml
# crate1
[features]
default = ["foo", "bar"]

# crate2
crate1 = { version = "...", default-features = false, features = ["bar"] } 
```

crates/client/Cargo.toml
calvinrp: Fixed

Cargo.toml
calvinrp: Ah, removing then.

calvinrp: Removed

Cargo.toml
calvinrp: Fixed

calvinrp:
Also, added a `output` option for the `warg download -o` subcommand. And asks the user if wants to have the file written to the CWD. Otherwise, just downloads to the local cache which is kind of confusing and hard to find.

crates/client/src/lib.rs
lann, line 1: Sorry I completely missed these changes in my first pass: "Large diffs are not rendered by default." :roll_eyes:

crates/client/src/lib.rs
```diff
@@ -61,30 +69,40 @@ where
     content: C,
     namespace_map: N,
     api: api::Client,
+    #[cfg(feature = "cli-interactive")]
```
lann, line 72: In general you're better off minimizing `#[cfg(feature ...)]` branches as much as possible, even if it means leaving (nominally) dead code around - these branches can end up being pretty fragile, especially if they aren't carefully tested (which itself isn't all that straightforward with rust/cargo).

As an overall strategy here I'd recommend leaning on the `disable_interactive` flag: force it to `true` when this feature is disabled and let that drive (most of) the rest of the conditional logic.

crates/client/src/lib.rs
```diff
@@ -61,30 +69,40 @@ where
...
     pub fn new(
         url: impl IntoUrl,
         registry: R,
         content: C,
         namespace_map: N,
         auth_token: Option<Secret<String>>,
-        ignore_federation_hints: bool,
-        auto_accept_federation_hints: bool,
+        #[cfg(feature = "cli-interactive")] ignore_federation_hints: bool,
+        #[cfg(feature = "cli-interactive")] auto_accept_federation_hints: bool,
```
lann, line 91: Don't these two flags still make sense without interaction?

crates/client/src/lib.rs
```diff
@@ -310,109 +328,173 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+        let mut init_record_id: Option<RecordId> = None;
+
+        let (package, record) = loop {
+            let mut info = publish_info.clone();
+
+            #[cfg(feature = "cli-interactive")]
+            let mut initializing = info.initializing();
+
+            #[cfg(not(feature = "cli-interactive"))]
+            let initializing = info.initializing();
```
lann, line 362: This sort of thing is definitely not worth feature-gating. If the compiler complains:

```suggestion
            #[allow(unused_mut)] // feature = "cli-interactive"
            let mut initializing = info.initializing();
```

crates/client/src/lib.rs
```diff
@@ -310,109 +328,173 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+                }) => {
+                    if !initializing && cfg!(not(feature = "cli-interactive")) {
                         return Err(ClientError::MustInitializePackage {
                             name,
                             has_auth_token,
                         });
+                    } else if !initializing {
+                        #[cfg(feature = "cli-interactive")]
+                        {
+                            if self.disable_interactive {
```
lann, line 390: You could probably move relevant `use`s down here, e.g.

```suggestion
                        #[cfg(feature = "cli-interactive")]
                        {
                            use dialoguer::{theme::ColorfulTheme, Confirm};
                            if self.disable_interactive {
```

crates/client/src/lib.rs
calvinrp: Revising

crates/client/src/lib.rs
calvinrp: Simplified

crates/client/src/lib.rs
calvinrp: fixed

crates/client/src/lib.rs
calvinrp: Simplified

crates/client/src/lib.rs
calvinrp: Fixed

crates/client/src/lib.rs
```diff
@@ -310,109 +316,164 @@ impl<R: RegistryStorage, C: ContentStorage, N: NamespaceMapStorage> Client<R, C,
...
+                                info.entries.insert(0, PublishEntry::Init);
+                                initializing = true;
+                                accepted_prompt_to_initialize = true;
+                            } else {
+                                return Err(ClientError::MustInitializePackage {
+                                    name,
+                                    has_auth_token,
+                                });
+                            }
+                        }
```
lann, line 397: What happens here when the feature is disabled? Should it be returning `MustInitializePackage`?

calvinrp: Revised to be cleaner and more correct.

lann (approved):

PR #275, opened 2024-05-09
moved the keyring functions back into warg_client

calvinrp:
This allows us to have a simpler `FileSystemClient::new_with_default_config()` that gets the expected `auth_token`.

Would simplify https://github.com/bytecodealliance/cargo-component/pull/288, https://github.com/bytecodealliance/wac/pull/103, https://github.com/bytecodealliance/wasm-pkg-tools/pull/6, https://github.com/esoterra/wow/pull/3

This:
```rust
        let config = warg_client::Config::from_default_file()?.unwrap_or_default();
        let url = url.or(config.home_url.as_deref());
        let auth_token = if config.keyring_auth && url.is_some() {
            get_auth_token(&RegistryUrl::new(url.unwrap())?)?
        } else {
            None
        };
        Ok(Self {
            client: FileSystemClient::new_with_config(url, &config, auth_token)?,
        })
```

would change to:

```rust
        Ok(Self {
            client: FileSystemClient::new_with_default_config(url)?,
        })
```


Cargo.toml
```diff
@@ -54,9 +54,10 @@ wit-parser = "0.13.1"
 testresult = "0.3.0"
 
 [features]
-default = ["cli-interactive"]
+default = ["cli-interactive", "keyring"]
 postgres = ["warg-server/postgres"]
 cli-interactive = ["warg-client/cli-interactive"]
+keyring = ["warg-client/keyring"]
```
lann, line 60: I don't know that we actually need to make this optional for the binary do we? Actually, I didn't really think about this until now: it probably doesn't make sense to have _any_ of these features; what does the `postgres` feature even mean here?

crates/client/src/lib.rs
```diff
@@ -1222,6 +1225,10 @@ impl FileSystemClient {
         let disable_interactive =
             cfg!(not(feature = "cli-interactive")) || config.disable_interactive;
 
+        if cfg!(feature = "keyring") && auth_token.is_none() && config.keyring_auth {
+            auth_token = crate::keyring::get_auth_token(&url)?
+        }
+
```
lann, line 1231: Does this compile with `cargo build --no-default-features`? We should probably add that as a check for CI...

lann (commented):
Partial review; I suspect this won't compile with `cargo build --no-default-features`

crates/client/src/lib.rs
calvinrp: It compiles for me.

Cargo.toml
calvinrp: The `postgres` feature is for the `warg-server` switches between using an in-memory vs Postgres backing when setting the feature flag. That probably still makes sense for now.

The other `keyring` and `cli-interactive` features might make sense to have the option of compiling without those implementations in the `warg-client`.

crates/client/src/lib.rs
lann:
```
$ cd crates/client && cargo build --no-default-features
...
error[E0433]: failed to resolve: could not find `keyring` in the crate root
```

Cargo.toml
lann: These features are just for the crate defined in this file: `warg-cli`

crates/client/src/lib.rs
calvinrp: Pushed fixed.

Cargo.toml
calvinrp: Ah, you mean we can drop `postgres` in the top-level Cargo.toml but keep it in the `crates/server/Cargo.toml`?

lann: Yeah, most of the root `Cargo.toml` is (somewhat confusingly) completely independent of other workspace members.

calvinrp: Removing `postgres` from the top-level `Cargo.toml` appears to affect how I run tests. It might affect the test runner, but would need to double check. But doesn't seem urgent.

lann: Ok; are the `keyring` and `cli-interactive` features needed?

lann (approved):

src/commands.rs
```diff
@@ -69,25 +67,18 @@ impl CommonOptions {
...
-            self.auth_token(config)?,
-        )? {
-            StorageLockResult::Acquired(client) => Ok(client),
-            StorageLockResult::NotAcquired(path) => {
-                println!(
-                    "blocking on lock for directory `{path}`...",
-                    path = path.display()
-                );
+        let client =
+            match FileSystemClient::try_new_with_config(self.registry.as_deref(), config, None)? {
```
dfoxfranke, line 71: Was it intentional that you're now passing `None` for the auth token (and that you deleted the now-unused `auth_token` method below)?

calvinrp: Yes, I was trying to simplify the burden on external crates getting the correct `auth_token`. So `None` gets the correct one, https://github.com/bytecodealliance/registry/blob/main/crates/client/src/lib.rs#L1403-L1405

Also, introduced `new_with_default_config()` to simplify, https://github.com/bytecodealliance/registry/blob/main/crates/client/src/lib.rs#L1428-L1430

PR #277, opened 2024-05-10
Clean up some clippy warnings

dfoxfranke:
Remove some unused imports, use `clone_from` where appropriate, allow unread fields on a Postgres-related structure, and fix a `#[cfg(feature = ...)]` referencing a non-existent feature.

### DIFF
--- a/crates/protocol/tests/operator.rs
+++ b/crates/protocol/tests/operator.rs
@@ -39,7 +39,7 @@ fn validate_input(input: Vec<EnvelopeData>) -> Result<LogState> {
             let key = signing::PrivateKey::decode(e_data.key.clone()).unwrap();
             let mut record: operator::OperatorRecord = e_data.contents.try_into().unwrap();
 
-            record.prev = last.clone();
+            record.prev.clone_from(last);
 
             let envelope = ProtoEnvelope::signed_contents(&key, record).unwrap();
 
--- a/crates/server/src/datastore/postgres/models.rs
+++ b/crates/server/src/datastore/postgres/models.rs
@@ -91,6 +91,7 @@ pub struct NewCheckpoint<'a> {
 
 #[derive(Queryable)]
 #[diesel(table_name = checkpoints)]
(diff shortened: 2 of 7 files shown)


crates/server/src/datastore/postgres/models.rs
```diff
@@ -91,6 +91,7 @@ pub struct NewCheckpoint<'a> {
 
 #[derive(Queryable)]
 #[diesel(table_name = checkpoints)]
+#[allow(dead_code)]
```
lann, line 94: What was marked dead here? I'm surprised anything in a `#[derive(Queryable)]` would be...

dfoxfranke: The timestamp fields. The query fills them in, but nothing reads them.

crates/protocol/tests/operator.rs
```diff
@@ -39,7 +39,7 @@ fn validate_input(input: Vec<EnvelopeData>) -> Result<LogState> {
             let key = signing::PrivateKey::decode(e_data.key.clone()).unwrap();
             let mut record: operator::OperatorRecord = e_data.contents.try_into().unwrap();
 
-            record.prev = last.clone();
+            record.prev.clone_from(last);
```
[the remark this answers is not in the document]
dfoxfranke: Want to disable the lint rather than make this change?

crates/server/src/datastore/postgres/models.rs
lann: Ah yeah I guess `Queryable` would be read-only wouldn't it? :+1:

crates/protocol/tests/operator.rs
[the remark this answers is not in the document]
lann: eh, given its just in tests its probably fine to leave this PR as-is.

lann (approved):

lann:
Thanks!

Issue #271, opened 2024-05-08
`linux-keyutils` credentials are lost after reboot


dfoxfranke:
#270 made `linux-keyutils` the default credential builder on Linux in order to resolve errors for users who do not have `gnome-keyring` installed. However, unlike `gnome-keyring`, `linux-keyutils` does not persist credentials on disk. They are stored in kernel memory only, are lost on reboot, and also time out after a configurable period of non-use. `linux-keyutils` is designed primarily around the needs of cloud-based virtual machine instances which themselves have no persistence and are provisioned at boot by some sort of key distribution server. It isn't a good fit for dev desktops, or at least not on its own.

There's no free lunch here, really. If credentials are going to survive across reboots, they need to be stored on disk somewhere. If they're stored in plaintext, that has the obvious security drawbacks, and there's no advantage to using the `keyring` crate at all rather than just using the filesystem. If they're stored encrypted, then the decryption key has to come from somewhere, so either you have to prompt for a password or it has to be already available from an agent, in which case the agent has to get it from somewhere. If you want decryption to appear seamless, then the only good option is to derive the key from the user's login password at the time of login, which is exactly what `gnome-keyring` does. But that can only work if the system is set up for it, having `gnome-keyring` installed and `pam_keyring` configured. So basically, you're going to have to choose between:

1. Storing keys in plaintext, or
2. Needing to prompt for a password at least sometimes,
3. Requiring some system infrastructure which isn't universally available,
4. Giving up on persistence altogether and requiring the user to regularly repeat the authorization workflow.

Ideally, you'd pick a sensible default (probably 3, maybe 2), but make it configurable.

calvinrp:
Release [v0.4.3](https://github.com/bytecodealliance/registry/releases/tag/v0.4.3) that reverts #270

calvinrp:
#287 and #278 were merged and should resolve. @dfoxfranke Thank you again

PR #278, opened 2024-05-10
Configurable keyring backends

dfoxfranke:
Makes keyring backends user-selectable, as discussed in https://github.com/bytecodealliance/registry/issues/271#issuecomment-2100908799.

- [x] Refactor `client::keyring` functions into a `Keyring` struct which can be instantiated with a backend selection.
- [x] Add a config option to control what backend is used.
- [x] Improve error reporting when the `secret-service` backend isn't working.
- [ ] Add a `flat-file` backend that stores keys in plaintext alongside the config file.

### DIFF
--- a/crates/client/src/keyring/mod.rs
+++ b/crates/client/src/keyring/mod.rs
@@ -2,22 +2,20 @@
 
 use crate::config::Config;
 use crate::RegistryUrl;
-use anyhow::{anyhow, bail, Context};
 use indexmap::IndexSet;
 use secrecy::Secret;
 use warg_crypto::signing::PrivateKey;
+
+mod error;
+use error::KeyringAction;
+pub use error::KeyringError;
 
 /// Interface to a pluggable keyring backend
 #[derive(Debug)]
 pub struct Keyring {
     imp: Box<keyring::CredentialBuilder>,
+    name: &'static str,
 }
-
-/// The type of keyring errors.
-///
-/// Currently just a synonym for [`anyhow::Error`], but will change to
-/// something capable of more user-friendly diagnostics.
-pub type KeyringError = anyhow::Error;
 
 /// Result type for keyring errors.
 pub type Result<T, E = KeyringError> = std::result::Result<T, E>;
@@ -108,8 +106,15 @@ impl Keyring {
     /// The argument should be an element of [Self::SUPPORTED_BACKENDS].
     pub fn new(backend: &str) -> Result<Self> {
         Self::load_backend(backend)
-            .ok_or_else(|| anyhow!("failed to initialize keyring: unsupported backend {backend}"))
-            .map(|imp| Self { imp })
+            .ok_or_else(|| KeyringError::unknown_backend(backend.to_string()))
+            .map(|imp| Self {
+                imp,
+                // Get an equivalent &'static str from our &str
+                name: Self::SUPPORTED_BACKENDS
+                    .iter()
+                    .find(|s| **s == backend)
+                    .expect("successfully-loaded backend should be found in SUPPORTED_BACKENDS"),
+            })
     }
 
     /// Instantiate a new keyring using the backend specified in a configuration file.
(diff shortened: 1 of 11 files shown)


dfoxfranke:
I'm marking this ready for review. The `flat-file` backend isn't implemented yet, but I think there's now enough in this PR to constitute an improvement for users, so it's suitable for merging.

crates/client/src/keyring/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+pub struct Keyring {
+    imp: Box<keyring::CredentialBuilder>,
+    name: &'static str,
+}
+
+/// Result type for keyring errors.
+pub type Result<T, E = KeyringError> = std::result::Result<T, E>;
+
+impl Keyring {
+    #[cfg(target_os = "linux")]
```
lann, line 24: FYI: https://crates.io/crates/match_cfg

crates/client/src/keyring/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+                    .expect("successfully-loaded backend should be found in SUPPORTED_BACKENDS"),
+            })
+    }
+
+    /// Instantiate a new keyring using the backend specified in a configuration file.
+    pub fn from_config(config: &Config) -> Result<Self> {
+        if let Some(ref backend) = config.keyring_backend {
+            Self::new(backend.as_str())
+        } else {
+            Self::new(Self::SUPPORTED_BACKENDS[0])
```
lann, line 125:
```suggestion
            Self::new(Self::DEFAULT_BACKEND)
```

lann:
Code lgtm; cc @calvinrp who has worked on keyring more recently

lann (commented):
Code lgtm; cc @calvinrp who has worked on keyring more recently

crates/client/src/keyring/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+        target_os = "windows",
+    )))]
+    /// List of supported credential store backends
+    pub const SUPPORTED_BACKENDS: &'static [&'static str] = &["mock"];
+
+    /// The default backend when no configuration option is set
+    pub const DEFAULT_BACKEND: &'static str = Self::SUPPORTED_BACKENDS[0];
+
+    /// Returns a human-readable description of a keyring backend.
+    pub fn describe_backend(backend: &str) -> &'static str {
```
macovedj, line 55: Would it be difficult to make `backend` an enum?

crates/client/src/keyring/mod.rs
dfoxfranke: I didn't think it worth pulling in the dependency just to use it in this one place where it barely provides any simplification. But if it gets pulled in later for wider use then, yes, this should be refactored.

crates/client/src/keyring/mod.rs
dfoxfranke: I considered this but decided it would create more problems than it would solve. The set of supported backends is platform-dependent, so there would be a distinction between the enumeration of all understood backend identifiers, and the subset of those which are supported on the user's platform. So there would be a design trade-off between:

1. Defining two separate enums, one which covers all backends and one which covers those which are supported on the platform.
2. Defining only the supported enum, and needing a bunch of conditional attributes every time you want to match on it.
3. Defining only the all-backends enum, and putting an `is_supported(&self) -> bool` method on it.

Then we have to decide how exactly we want our `Serialize` and `Deserialize` implementations to work and which type they should use, and probably put custom error reporting into this traits to not be too cryptic if the user copies a configuration file from one system to another where the backend isn't supported. Altogether it didn't seem worth it, so I just kept this stringly typed.

calvinrp:
Thank you! This looks great.

calvinrp (approved):

crates/client/src/keyring/mod.rs
dfoxfranke: Calvin already merged, so I'll fix this in my next PR.

PR #287, opened 2024-05-15
Implement flat-file keyring backend

dfoxfranke:
This adds support for storing warg credentials as flat files in the warg config directory, following up on #278.


calvinrp (approved):
Thank you!

PR #291, opened 2024-05-16
Well known config

calvinrp:
- Added Warg client support for `.well-known` path, which required the `warg_client::FileSystemClient::new_with_config()` methods to be `async`;
- Support both `warg login <registry-url>` and `warg login --registry <registry-url>` (also for `warg logout`);
- Sets the default registry, without configuration, to be `bytecodealliance.org`; [See discussions](https://github.com/bytecodealliance/wasm-pkg-tools/discussions/3#discussioncomment-9359731)

Follow on work to respect HTTP cache headers.

### DIFF
--- a/crates/client/src/api.rs
+++ b/crates/client/src/api.rs
@@ -12,20 +12,24 @@ use secrecy::{ExposeSecret, Secret};
 use serde::de::DeserializeOwned;
 use std::borrow::Cow;
 use thiserror::Error;
-use warg_api::v1::{
-    content::{ContentError, ContentSourcesResponse},
-    fetch::{
-        FetchError, FetchLogsRequest, FetchLogsResponse, FetchPackageNamesRequest,
-        FetchPackageNamesResponse,
+use warg_api::{
+    v1::{
+        content::{ContentError, ContentSourcesResponse},
+        fetch::{
+            FetchError, FetchLogsRequest, FetchLogsResponse, FetchPackageNamesRequest,
+            FetchPackageNamesResponse,
+        },
+        ledger::{LedgerError, LedgerSourcesResponse},
+        monitor::{CheckpointVerificationResponse, MonitorError},
+        package::{ContentSource, PackageError, PackageRecord, PublishRecordRequest},
+        paths,
+        proof::{
+            ConsistencyRequest, ConsistencyResponse, InclusionRequest, InclusionResponse,
+            ProofError,
+        },
+        REGISTRY_HEADER_NAME, REGISTRY_HINT_HEADER_NAME,
     },
-    ledger::{LedgerError, LedgerSourcesResponse},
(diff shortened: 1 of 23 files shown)


crates/client/src/api.rs
```diff
@@ -216,6 +223,31 @@ impl Client {
     pub fn url(&self) -> &RegistryUrl {
         &self.url
     }
+    /// Gets the `.well-known` configuration registry URL.
+    pub async fn well_known_config(&self) -> Result<Option<RegistryUrl>, ClientError> {
+        let url = self.url.join(WELL_KNOWN_PATH);
+        tracing::debug!(url, "getting `.well-known` config",);
+
+        let res = self.client.get(url).send().await?;
+
```
lann, line 232: Maybe `tracing::debug` the response status here? Or at least in the `!is_success` case below.

macovedj (approved):

src/commands/logout.rs
```diff
@@ -10,11 +10,21 @@ pub struct LogoutCommand {
     /// The common command options.
     #[clap(flatten)]
     pub common: CommonOptions,
+
+    /// The URL of the registry to use.
```
lann, line 14: Should we call out that this is an advanced option and plain `registry` is preferred?

calvinrp: How do you mean "call out"?

lann: Maybe I'm just confused; what is the purpose of these two different arguments?

crates/client/src/api.rs
```diff
@@ -230,20 +230,23 @@ impl Client {
...
-                    ClientError::InvalidWellKnownConfig(self.url.registry_domain().to_string())
-                })?
-                .warg_url
-            {
-                Ok(Some(RegistryUrl::new(warg_url)?))
-            } else {
-                Ok(None)
-            }
+        if !res.status().is_success() {
+            tracing::debug!("the `.well-known` config was not found");
```
lann, line 234: Can we include the actual `res.status()` here? I'm mostly interested in this logging for surprising non-404 failures.

src/commands/logout.rs
calvinrp: I was trying to support both `warg login --registry <registry-url>` (to break existing instructions but also migrate to `warg login <registry-url>`.

Been trying, but `clap` doesn't really support it best I can tell with the `derive` args

crates/client/src/api.rs
calvinrp: Fixed

src/commands/logout.rs
lann: Ah OK; maybe pick one of the two to `#[arg(hide)]` and comment the reasoning?

calvinrp: Fixed

PR #298, opened 2024-08-30
added rustls-tls feature

calvinrp:


crates/client/Cargo.toml
```diff
@@ -10,8 +10,9 @@ homepage = { workspace = true }
 repository = { workspace = true}
 
 [features]
-default = ["cli-interactive", "keyring"]
+default = ["cli-interactive", "keyring", "rustls-tls"]
 native-tls-vendored = ["reqwest/native-tls-vendored"]
+rustls-tls = ["reqwest/rustls-tls"]
```
thomastaylor312, line 15: You'll need to turn off default features in reqwest and then have a feature that enables native-tls in reqwest and then a rustls-tls feature for enabling that (like you have now)

thomastaylor312 (changes requested):

thomastaylor312 (approved):

PR #301, opened 2024-09-24
fix missing arg in README command

MarinPostma:
The getting started in the readme is missing an argument :)


calvinrp (approved):
Thank you!

PR #313, opened 2025-01-20
chore(ci): Update wasm32-wasi reference for Rust 1.84

joonas:
Per the failure in https://github.com/bytecodealliance/registry/pull/312, Rust 1.84 deprecated `wasm32-wasi` target, and thus it now needs to be set to either `wasm32-wasip1` or `wasm32-wasip2`.

I wasn't sure if we wanted to target `p1` or `p2`, so I defaulted to `p2`, but happy to switch it to `p1` if that's preferred.


joonas:
Going to fold these into #312.

